PR #3, opened 2017-10-10
derive constants from BN parameter

cdetrio:
Its a bit more intuitive to derive the various constants from the minimal set of magic numbers.


pipermerriam:
@cdetrio I'm not really qualified to judge the correctness of this PR but if you'd still like to merge these changes what do you think about:

1. Rebasing and fixing the conflicts
2. Adding some basic test cases (assuming there are some worth adding)

Issue #11, opened 2018-10-17
Add type hinting


pipermerriam:
> Largely copy/pasted from https://github.com/ethereum/py-evm/issues/1398

### Background

[Type hints](https://www.python.org/dev/peps/pep-0484/) allow us to perform static type checking, among other things. They raise the security bar by catching bugs at development time that, without type support, may turn into runtime bugs.

[This stackoverflow answer](https://stackoverflow.com/questions/32557920/what-are-type-hints-in-python-3-5/32558710#32558710) does a great job at describing their main benefits.

### What is wrong?

This library currently does not have any type hints.

This needs to be fixed by:

1. Adding all missing type hints.
2. Enforcing (stricter) type checking in CI

### How

There does exist tooling ([monkeytype](https://engineering.instagram.com/let-your-code-type-hint-itself-introducing-open-source-monkeytype-a855c7284881)) to the generation of type hints for existing code bases. From my personal experience `monkeytype` *can* be helpful but does still require manual fine tuning. Also, manually adding these type hints does serve as a great boost to the general understanding of the code base as it forces one to think about the code.

1. Run `mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth`

2. Eliminate every reported error by adding the right type hint

Because this library supports older versions of python, the type hints will not be able to use the modern python3.6 syntax.

### Definition of done

This issue is done when the following criteria are met:

1. `mypy` is run in CI

Add a new command to the `flake8` environment in the `tox.ini` file that runs:

```
mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p py_ecc`
```

2. Usage of `type: ignore` (silencing the type checker) is minimized and there's a reasonable explanation for its usage 

### Stretch goals

When this issue is done, stretch goals can be applied (and individually get funded) to tighten type support to qualify:

1. `mypy --strict --follow-imports=silent --ignore-missing-imports --no-strict-optional -p py_ecc`

6ug:
I like to work on this thing, I just saw talk from guido and I can do this thing.

pipermerriam:
@6ug :+1: 

[...]

PR #18, opened 2018-10-18
[WIP] Added type hinting

6ug:
Fixes https://github.com/ethereum/py_ecc/issues/11
Totally fine tuned manually! automated tools not handling dynamic execution.

### DIFF
--- a/py_ecc/bn128/bn128_curve.py
+++ b/py_ecc/bn128/bn128_curve.py
@@ -6,7 +6,7 @@ from .bn128_field_elements import (
     FQ2,
     FQ12,
 )
-from typing import Tuple, Union
+from typing import Tuple
 
 
 curve_order = 21888242871839275222246405745257275088548364400416034343698204186575808495617
--- a/py_ecc/bn128/bn128_curve.py
+++ b/py_ecc/bn128/bn128_curve.py
@@ -6,6 +6,7 @@ from .bn128_field_elements import (
     FQ2,
     FQ12,
 )
+from typing import Tuple, Union
 
 
 curve_order = 21888242871839275222246405745257275088548364400416034343698204186575808495617
@@ -42,7 +43,7 @@ Z2 = None
 
 
 # Check if a point is the point at infinity
-def is_inf(pt):
+def is_inf(pt: Tuple[int]) -> bool:
     return pt is None
 
 
--- a/py_ecc/bn128/bn128_field_elements.py
+++ b/py_ecc/bn128/bn128_field_elements.py
@@ -1,7 +1,8 @@
 from __future__ import absolute_import
 
 import sys
-
+from typing import NewType, Tuple
+from py_ecc.types import Point2D, Point3D
 
 sys.setrecursionlimit(10000)
 
@@ -11,7 +12,6 @@ if sys.version_info.major == 2:
     int_types = (int, long)  # noqa: F821
 else:
     int_types = (int,)
-
 
 # The prime modulus of the field
 field_modulus = 21888242871839275222246405745257275088696311157297823662689037894645226208583
@@ -24,7 +24,7 @@ FQ12_modulus_coeffs = [82, 0, 0, 0, 0, 0, -18, 0, 0, 0, 0, 0]  # Implied + [1]
 
 # Extended euclidean algorithm to find modular inverses for
 # integers
-def inv(a, n):
+def inv(a: int, n: int) -> int:
     if a == 0:
         return 0
     lm, hm = 1, 0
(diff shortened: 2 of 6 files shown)


py_ecc/bn128/bn128_field_elements.py
```diff
@@ -119,14 +119,14 @@ def zero(cls):
 
 
 # Utility methods for polynomial math
-def deg(p):
+def deg(p: Tuple[int, int, int]) -> int:
```
pipermerriam: It may be prudent for us to define a new `Point` type for this to add an additional level of type safety.

```python
Point = typing.NewType('Point', Tuple[int, int, int])
```

This prevents passing just *any* `(int, int, int)` into these functions, requiring us to be explicit about our point abstraction.

And looking at the code, it appears we'd need both a `Point2D` and a `Point3D` to handle the different dimensionalities that are used.

py_ecc/secp256k1/secp256k1.py
```diff
@@ -137,7 +137,7 @@ def ecdsa_raw_sign(msghash, priv):
     return v, r, s
 
 
-def ecdsa_raw_recover(msghash, vrs):
+def ecdsa_raw_recover(msghash: str, vrs: Tuple[int, int, int]) -> Union[bool, Tuple[int, int]]:
```
pipermerriam: It'd be nice to refactor this so to replace `return False` with an exception so that the return type can be more firm here.  Is that something you'd feel comfortable doing?

6ug: sure @pipermerriam, I like to add 
```
if (xcubedaxb - y * y) % P != 0 or not (r % N) or not (s % N):
	return ValueError("sig is invalid, %d cannot be the x coord for point on curve" % r)
```

py_ecc/bn128/bn128_curve.py
```diff
@@ -41,7 +42,7 @@
 Z2 = None
 
 # Check if a point is the point at infinity
-def is_inf(pt):
```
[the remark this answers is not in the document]
6ug: @pipermerriam  can you explain little-bit here, I mean it is already `Tuple`.

py_ecc/bn128/bn128_field_elements.py
6ug: @pipermerriam can you check this changes https://github.com/ethereum/py_ecc/pull/18/commits/e4de2beb46d13c04f0dbd8aae46c681e7a7a2f73

py_ecc/bn128/bn128_curve.py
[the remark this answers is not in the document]
pipermerriam: I 1) commented on the wrong section of code and 2) I made my comment earlier when this was a `List` type.  Disregard.

py_ecc/secp256k1/secp256k1.py
```diff
@@ -12,6 +12,8 @@ def safe_ord(value):
         else:
             return ord(value)
 
+Point2D = Tuple[int, int]
+Point3D = Tuple[int, int, int]
```
pipermerriam: These could go in a common `py_ecc.types` module so they don't have to be repeated.

tox.ini
```diff
@@ -20,9 +20,12 @@ basepython =
...
     py36: python3.6
+    py37: python3.7
+
 
 [testenv:lint]
 basepython=python
 extras=lint
 commands=
     flake8 {toxinidir}/py_ecc
+    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p py_ecc
```
pipermerriam: just noting my copy/paste error, you can drop the `-p p2p` from this command.  Also, I'm going to advocate that this be updated to the stretch goals command since this repo should be as strictly typed as possible given its cryptographic nature.

```
mypy --strict --follow-imports=silent --ignore-missing-imports --no-strict-optional -p py_ecc
```

6ug:
PR https://github.com/ethereum/py_ecc/pull/19 should fix ^^ above CI failures.

tox.ini
6ug: Opps, my bad, didn't noticed that, correcting.

py_ecc/secp256k1/secp256k1.py
6ug: @pipermerriam Great idea indeed, updated with changes! :+1:

pipermerriam:
Looks like ci failures are some small linting errors, and failing due to python 2.7 not supporting type hints.

@carver I'm inclined to drop python 2 support.  Can you confirm?

carver:
> I'm inclined to drop python 2 support. Can you confirm?

Yeah, the next release can be v2 :+1:

pipermerriam:
@6ug looks like this needs to be rebased against `master`

6ug:
@pipermerriam our last `typing: Module not found` issue was due to - python3.4 does not have inbuilt support for [PEP-484](https://www.python.org/dev/peps/pep-0484/) type hinting, we need to externally add `typing` in this case, https://github.com/ethereum/py_ecc/pull/18/commits/d86ca0e2afe7a69cc4b6b967bc81ba0b03a2db90 will fix that.

carver:
We should drop py3.4 as well, while we're dropping old python versions. Maybe even py3.5 too.

pipermerriam:
ping @6ug are you still working on this?

pacrob:
Closed by #143

Issue #10, opened 2018-10-17
Support for python 3.7


pipermerriam:
Add support for python 3.7

- Add test run to `tox.ini`
- Add test run to `.travis.yml`
- Fix any issues that show up in tests.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 90.0 DAI (90.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/py_ecc/10/1523).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/py_ecc/10/1523).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $48,158.14 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 2 months, 4 weeks from now.
Please review their action plans below:


**1) [FabioRosado](https://gitcoin.co/profile/FabioRosado) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/py_ecc/10/1523?mutate_worker_action=approve&worker=FabioRosado) | [reject worker](https://gitcoin.co/issue/ethereum/py_ecc/10/1523?mutate_worker_action=reject&worker=FabioRosado))_.**

Hello,
I am interested in working in this issue. 

The first thing to do will be to add python3.7 to tox and travis, then run the tests and try to fix the issues encountered there one by one.

This is my first time doing something like this, hopefully I can do a good job :)

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/py_ecc/10/1523).


[...]

PR #19, opened 2018-10-18
Adding python3.7 support 

6ug:
Fixes https://github.com/ethereum/py_ecc/issues/10

(PS. I need this PR to be get merge to proceed with my mypy hinting issue.)
Not able to apply from site due to it thrown me some bad API error.


carver:
This should be using CircleCI instead of adding a travis config.

6ug:
Things I did - 
1. Configured `circleci` for `python37`
2. Configured `tox` for `python37`
3. Updated `classifier`
4. Added 100 chars lints limit (extra thing)
5. Added `flake8` requirement (extra thing)

@carver kindly review extra things and let me know, may be I can remove them upon suggestions.

Makefile
```diff
@@ -22,7 +22,7 @@ clean-pyc:
 	find . -name '*~' -exec rm -f {} +
 
 lint:
-	flake8 py_ecc
+	flake8 --max-line-length=100 py_ecc
```
carver: Instead, this can go in the tox.ini, like https://github.com/ethereum/ethereum-python-project-template/blob/03f00d3fbe73b0ea6bd46b5e5b4e9a4103ee0226/tox.ini#L17-L20

requirements-dev.txt
```diff
@@ -1,2 +1,3 @@
 pytest==3.2.2
 pytest-xdist==1.20.1
+flake8==3.5.0
```
carver: The flake8 dependency is in the `py_ecc[lint]` extra, we should be using that instead.
In fact, requirements-dev and any references to it should be removed entirely.

Makefile
6ug: yes @carver, then I like to go with earlier one so removing this.

carver: > do you mind if I ignore it from commit?

Sorry, do you mean to rebase and force-push? That's totally fine.

tox.ini
```diff
@@ -14,7 +14,9 @@ commands=py.test {posargs:tests}
 passenv=
     PYTEST_ADDOPTS
 deps =
-    -r{toxinidir}/requirements-dev.txt
```
6ug: I tried to remove this `requirement` and added dependency under `deps` of `tox`. I tried lots of options here but failed so this is simplest option I got. @carver calling you for wisdom!

carver: Use `extras=test` and you can drop all these `deps`

6ug: yup, worked like a charm, thanks @carver ! :100:

tox.ini
```diff
@@ -14,15 +14,16 @@ commands=py.test {posargs:tests}
...
     py35: python3.5
     py36: python3.6
+    py37: python3.7
 
 [testenv:lint]
 basepython=python
 extras=lint
 commands=
-    flake8 {toxinidir}/py_ecc
+    flake8 --max-line-length=100 {toxinidir}/py_ecc
```
carver:
```suggestion
    flake8 {toxinidir}/py_ecc
```

Shouldn't be necessary ^

6ug: Agreed!

Issue #12, opened 2018-10-17
Expose type hints via PEP561


pipermerriam:
blocked until #11 is complete

## What was wrong?

[PEP561](https://www.python.org/dev/peps/pep-0561/) allows us to expose type hints from this library to other libraries which use it.  Without this, any library which uses `py_ecc` and `mypy` style type checking doesn't benefit from the additional type safety.

## Definition of done?

Do what was done in https://github.com/ethereum/eth-typing/pull/10 to allow this data to be exposed when this library is installed as a package.

6ug:
@pipermerriam is this eligible to bounty?

PR #20, opened 2018-10-18
[RFR] Enabling type hinting, applied PEP561

6ug:
Fixes https://github.com/ethereum/py_ecc/issues/12, Requesting for review! :100:


6ug:
Requesting for review! :100:

PR #22, opened 2018-10-28
Drop python 3.4 and 3.5 support

6ug:
Part of https://github.com/ethereum/py_ecc/issues/11 fixes, raised after https://github.com/ethereum/py_ecc/pull/18#issuecomment-433678030

Tested against code, works well!
screenshot from 2018-10-28 13-17-26

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,15 +38,13 @@ setup(
     package_data={'py_ecc': ['py.typed']},
     install_requires=[
     ],
-    python_requires='>=3.4, <4',
+    python_requires='>=3.6, <4',
     extras_require=extras_require,
     classifiers=[
         'Development Status :: 3 - Alpha',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
-        'Programming Language :: Python :: 3.4',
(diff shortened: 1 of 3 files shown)


6ug:
@carver @pipermerriam please review! :+1:

setup.py
```diff
@@ -38,15 +38,13 @@
...
-    python_requires='>=3.4, <4',
+    python_requires='>=3.6, <4',
     extras_require=extras_require,
     classifiers=[
         'Development Status :: 3 - Alpha',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
-        'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5',
```
carver: Cool, thanks for keeping all the extra stuff up to date, too.

pipermerriam:
I think we need to keep 3.5 till py-evm drops it...

https://github.com/ethereum/py-evm/blob/9870841a126fb53a6cbab2ef7bf3d46262fa23a8/setup.py#L16

but we can just do this as a major version bump so lets just do that.

PR #23, opened 2018-11-28
Type Hinting For py_ecc module

### LINKED ISSUE #11: Add type hinting

Bhargavasomu:
Fixes #11

### DIFF
--- a/py_ecc/bn128/bn128_field_elements.py
+++ b/py_ecc/bn128/bn128_field_elements.py
@@ -2,15 +2,15 @@ from __future__ import absolute_import
 
 import sys
 
+from typing import (
+    cast,
+    List,
+    Sequence,
+    Union,
+)
+
 
 sys.setrecursionlimit(10000)
-
-
-# python3 compatibility
-if sys.version_info.major == 2:
-    int_types = (int, long)  # noqa: F821
-else:
-    int_types = (int,)
 
 
 # The prime modulus of the field
@@ -19,12 +19,13 @@ field_modulus = 2188824287183927522224640574525727508869631115729782366268903789
 assert pow(2, field_modulus, field_modulus) == 2
 
 # The modulus of the polynomial in this representation of FQ12
-FQ12_modulus_coeffs = [82, 0, 0, 0, 0, 0, -18, 0, 0, 0, 0, 0]  # Implied + [1]
+FQ12_MODULUS_COEFFS = [82, 0, 0, 0, 0, 0, -18, 0, 0, 0, 0, 0]  # Implied + [1]
+FQ2_MODULUS_COEFFS = [1, 0]
 
 
 # Extended euclidean algorithm to find modular inverses for
 # integers
-def inv(a, n):
+def inv(a: int, n: int) -> int:
     if a == 0:
         return 0
     lm, hm = 1, 0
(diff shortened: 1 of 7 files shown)


Bhargavasomu:
@pipermerriam I have added the type hinting for the `bn128` and `secp256k1` module. Please review this and if everything is ok, I will also make the same type hinting to the `optimized_bn128` module, as it is similar to `bn128`.

vbuterin:
Looks good to me!

py_ecc/bn128/bn128_field_elements.py
```diff
@@ -39,52 +40,54 @@ def inv(a, n):
 # A class for field elements in FQ. Wrap a number in this class,
 # and it becomes a field element.
 class FQ(object):
```
pipermerriam: I'm wondering if this shouldn't inherit from `int` or `numbers.Number`.  I think that it might simplify the typing (maybe) since each of the methods argument types that are currently `Union[int, "FQ")` could be reduced to just be `int`.

Bhargavasomu: @pipermerriam that would involve refactoring the whole codebase, and maybe we can open a new issue once we are done with this?

PR #26, opened 2018-11-28
Ensure recursion limit is not reduced

pipermerriam:
### What was wrong

The way the recurrsion limit was being set it could end up lowing the limit.

### How was it fixed

Added a `max` to the logic to ensure it doesn't lower the current limit.


pipermerriam:
replaced by https://github.com/ethereum/py_ecc/pull/28

PR #28, opened 2018-11-28
Piper/cleanup linting recursion and mutability

pipermerriam:
### What this does

This fixes 3 things at once.

- ensure recursion limit setting doesn't reduce recursion limit
- convert `FQ` classes to use immutable data structures
- linting issues


py_ecc/optimized_bls12_381/optimized_field_elements.py
```diff
@@ -36,7 +28,7 @@ def __init__(self, n):
             self.n = n.n
         else:
             self.n = n % field_modulus
-        assert isinstance(self.n, int_types)
+        assert isinstance(self.n, int)
```
carver: Not sure it's worth double-checking, but thought I would add the reminder that `isinstance(bool(), int) == True`

carver:
Oh I see py2 support was already dropped, this just removes some cruft.

Bhargavasomu:
@pipermerriam could you merge this so that I can rebase on top of this, for closing #29?

pipermerriam:
@Bhargavasomu done.

Issue #31, opened 2018-12-06
Library refactor


pipermerriam:
By @Bhargavasomu from: https://github.com/ethereum/py_ecc/issues/24#issuecomment-442701351

The following things come to my mind.
* The classes `FQ`, `FQP`, `FQ2` and `FQ12` need not be reinitialized every time as they are not dependent on the type of curve or extension. So probably we could have these created in `field_elements.py` and we could use them everywhere (`bn128`, `optimized_bn128`, `bls`, `optimized_bls`).
* We could also have a general class `BaseCurve`, and then maybe every curve (`bn128_curve`, `optimized_bn128_curve`, ...) could inherit this and make the changes specific to the inherited class.
* We should probably move the constants into a seperate file (`constants.py`)
* We should also remove the assert statements which are not part of any function, but are part of the script in general, as shown
https://github.com/ethereum/py_ecc/blob/067a40261ad39526f6aa8bd69def7d6982993d13/py_ecc/bn128/bn128_pairing.py#L74-L83
* Also the type hinting should be further generalized wherever possible (in terms of removing redundant types; like `Optimized_FQPoint2D` could be replaced by `FQPoint2D`). Similary the type hinting should be carried out for the `bs12_381` and `optimized_bs12_381` submodules.

Also the only difference I see in all the curves is 
* Difference in the constants such as `b`, `b2`, `b12`, `G2`, `G12` ...
* Difference in the `twist` function

@vbuterin are my facts right or am I missing anything. 
@pipermerriam is the above design ok?

pipermerriam:
@Bhargavasomu I've submitted a request to have a bounty attached to this.

Bhargavasomu:
On this.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 250.0 DAI (250.0 USD @ $1.0/DAI)  attached to it as part of the Ethereum Foundation fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/py_ecc/31/1924).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/py_ecc/31/1924).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
[...]

Issue #24, opened 2018-11-28
Generalize fields and curves


vbuterin:
Currently, there are separate folders and files for bn128 and bls-12-381, with ~90% duplicated code.

It should be possible to generalize this more, creating a generic prime field class much like there is a semi-generic FQP class, as well as generic elliptic curves, and then only have separate files for the two existing curve types to define curve parameters and twist formulas and the slightly different pairing functions.

Bhargavasomu:
@vbuterin I am interested in generalizing and making the code flexible, could you please tell me the source of this whole codebase, such as a research paper or something.

vbuterin:
Elliptic curve pairings in general:

Bhargavasomu:
The following things come to my mind.
* The classes `FQ`, `FQP`, `FQ2` and `FQ12` need not be reinitialized every time as they are not dependent on the type of curve or extension. So probably we could have these created in `field_elements.py` and we could use them everywhere (`bn128`, `optimized_bn128`, `bls`, `optimized_bls`).
* We could also have a general class `BaseCurve`, and then maybe every curve (`bn128_curve`, `optimized_bn128_curve`, ...) could inherit this and make the changes specific to the inherited class.
* We should probably move the constants into a seperate file (`constants.py`)
* We should also remove the assert statements which are not part of any function, but are part of the script in general, as shown
https://github.com/ethereum/py_ecc/blob/067a40261ad39526f6aa8bd69def7d6982993d13/py_ecc/bn128/bn128_pairing.py#L74-L83

Also the only difference I see in all the curves is 
* Difference in the constants such as `b`, `b2`, `b12`, `G2`, `G12` ...
* Difference in the `twist` function

@vbuterin are my facts right or am I missing anything. 
@pipermerriam is the above design ok?

pipermerriam:
Design seems :+1: 

Removing all of the runtime assertions would be nice, but lets make sure that they are all still executed as part of the test suite.

Bhargavasomu:
@vbuterin in the `twist` function of `bn128` and `optimized_bn128`, we are multiplying `x` by `w^2` and `y` by `w^3`, whereas we should actually be dividing right? I think this is a bug (please correct me if I am wrong). Could you please follow up on this? (We are doing correctly by dividing in the case of `bls_12_381` curve). Attaching the snippet for reference.
[...]

PR #35, opened 2018-12-10
Refactoring in Naive Form

Bhargavasomu:
### What was wrong?
Fixes Issue: #31 

#### Cute Animal Picture


Bhargavasomu:
@pipermerriam I have not yet refactored the `pairing` part. But I have refactored the `field_elements`(so that all the curves can use the same code) and the `curve` implementation with the `BaseCurve` class implementation. 
* Please let me know if any changes are needed as far as design is concerned.
* The other main thing that depresses me is that, I haven't changed the `logic` anywhere, but just because of the usage of classes and inheritance, the performance got worser for a test case. (Tested in `ipython`). Attaching the test case below.
```
In [2]: from py_ecc import bn128

In [3]: time bn128.multiply(bn128.G12, 100)
CPU times: user 68.3 ms, sys: 0 ns, total: 68.3 ms
Wall time: 67.5 ms
Out[3]: 
((0, 0, 6289622313332368275808678712703019680903267734760705664249053540194934405010, 0, 0, 0, 0, 0, 19765057837394621075835443081152314037672821821523780638388756299265215370750, 0, 0, 0),
 (0, 0, 0, 15824998746886404237735115559171157131044885936835913698791405128162671676610, 0, 0, 0, 0, 0, 10168740664245831497642710459131680020400674013216613216484698513127357137715, 0, 0))

In [4]: a1 = BN128_Curve()

In [5]: time a1.multiply(a1.G12, 100)
CPU times: user 5.27 s, sys: 7.1 ms, total: 5.27 s
Wall time: 5.27 s
Out[5]: 
((0, 0, 6289622313332368275808678712703019680903267734760705664249053540194934405010, 0, 0, 0, 0, 0, 19765057837394621075835443081152314037672821821523780638388756299265215370750, 0, 0, 0),
 (0, 0, 0, 15824998746886404237735115559171157131044885936835913698791405128162671676610, 0, 0, 0, 0, 0, 10168740664245831497642710459131680020400674013216613216484698513127357137715, 0, 0))

```
I could really use your help and thoughts for this. (Been stuck on this for the whole day)

py_ecc/BaseCurve.py
```diff
@@ -0,0 +1,291 @@
...
+    # Generator for twisted curve over FQ12
+    G12 = None
+    # Point at infinity over FQ
+    Z1 = None
+    # Point at infinity for twisted curve over FQ2
+    Z2 = None
+
+    def __init__(self, curve_properties, curve_name):
+        self.curve_name = curve_name
+        self.curve_order = curve_properties[curve_name]["curve_order"]
```
pipermerriam: It appears that we only access `curve_properties[curve_name]` which suggests that we could modify the `__init__` argument to just take that value so that these lookups become:

```python
self.b = curve_properties["b"]
self.b2 = curve_properties["b"]
```

py_ecc/BaseCurve.py
```diff
@@ -0,0 +1,291 @@
...
+        self.b = curve_properties[curve_name]["b"]
+        self.b2 = curve_properties[curve_name]["b2"]
+        self.b12 = curve_properties[curve_name]["b12"]
+        self.G1 = curve_properties[curve_name]["G1"]
+        self.G2 = curve_properties[curve_name]["G2"]
+        self.G12 = self.twist(cast(Point2D[FQP], self.G2))
+        self.Z1 = curve_properties[curve_name]["Z1"]
+        self.Z2 = curve_properties[curve_name]["Z2"]
+
+    def is_inf(self, pt: GeneralPoint[Field]) -> bool:
```
pipermerriam: This can be updated to be a `staticmethod`

py_ecc/BaseCurve.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        Check if a point is the point at infinity
+        """
+        return pt is None
+
+    def is_on_curve(self, pt: Point2D[Field], b: Field) -> bool:
```
pipermerriam: This can be changed to a `classmethod`

py_ecc/BaseCurve.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        Check that a point is on the curve
+        """
+        if self.is_inf(pt):
+            return True
+        x, y = pt
+        return y**2 == x**3 + b
+
+    def double(self, pt: Point2D[Field]) -> Point2D[Field]:
```
pipermerriam: This can be updated to be a `staticmethod`

py_ecc/BaseCurve.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        Elliptic Curve Doubling (P+P).
+        """
+        x, y = pt
+        m = (3 * x**2) / (2 * y)
+        newx = m**2 - 2 * x
+        newy = -m * newx + m * x - y
+        return (newx, newy)
+
+    def add(self,
```
pipermerriam: can be a `classmethod`

py_ecc/BaseCurve.py
```diff
@@ -0,0 +1,291 @@
...
+    G12 = None
+    # Point at infinity over FQ
+    Z1 = None
+    # Point at infinity for twisted curve over FQ2
+    Z2 = None
+
+    def __init__(self, curve_properties, curve_name):
+        self.curve_name = curve_name
+        self.curve_order = curve_properties[curve_name]["curve_order"]
+        self.b = curve_properties[curve_name]["b"]
```
pipermerriam: All of these look like they'd be better done as properties of the class itself rather than needing to instantiate the class to get access to them.  Then we can define a subclass of the `BaseCurve` class for each curve we support so that we don't have to manage instantiating each one?

py_ecc/curve_properties.py
```diff
@@ -0,0 +1,163 @@
...
+    FQ2 as optimized_FQ2,
+    FQ12 as optimized_FQ12,
+)
+
+from py_ecc.new_typing import (
+    Point2D,
+)
+
+
+curve_properties = {
```
pipermerriam: I think that this would be better done using a `typing.NamedTuple` so that each of these properties can have stronger types.

py_ecc/field_elements.py
```diff
@@ -40,21 +23,33 @@ def inv(a: int, n: int) -> int:
...
+    def __init__(self, val: IntOrFQ, curve_name: str) -> None:
+        """
+        curve_name can be either 'bn128' or 'bls12_381'
+        This is needed to obtain field_modulus, FQ2_MODULUS_COEFFS
+        and FQ12_MODULUS_COEFFS from the curve properties
+        """
+        self.curve_name = curve_name
+        self.field_modulus = field_properties[curve_name]["field_modulus"]
+        # See, it's prime!
+        assert pow(2, self.field_modulus, self.field_modulus) == 2
```
pipermerriam: Runtime assertions are something we try to avoid because they are easily disabled.  Can you convert this to raise a `ValueError`?

pipermerriam:
@Bhargavasomu some preliminary review.

I don't think that the performance hit you're seeing is something we'll be able to merge so that will need to be addressed. It doesn't have to be equally fast, but it needs to be in the same order of magnitude.

Bhargavasomu:
@pipermerriam found out the devil causing more time and have fixed that. Just waiting for a reply from vitalik as part of https://github.com/ethereum/py_ecc/issues/24#issuecomment-446088199. Will make the next PR after that.

But one more thing that I noticed is that, after we refactor this, we would need to change the code in 
`py-evm` as well. For example, we have many places where `bn128.FQ2` is used. But as part of this refactor `FQ2` or any `FQ*` for that matter will not be binded to the curve anymore(like `bn128` or `bls12_381` etc.). So such changes have to be made in `py-evm` as well. My doubt was that if we merged this without changing the 
`py-evm` code which uses this library, then the tests would crash( if at all we test the `precompiles` as of now). So how do we go about this?

pipermerriam:
@Bhargavasomu I think we'll need to cut a v2 release to account for the breaking API change and more formally, we need to establish some documentation and boundaries for public/private API since currently everything in the library is effectively public API.

Bhargavasomu:
@pipermerriam the following issues still have not been addressed till this point and I have some questions.
* Instead of using `classmethod` or `staticmethod` decorators, I was thinking that maybe we could instantiate an object in `_init__.py` like `bn128 = BN128_Curve()`, and the users could use this object itself directly. Need your suggestion on this.
* I still haven't received a response to the comments raised by me [here](https://github.com/ethereum/py_ecc/issues/24#issuecomment-446088199), and hence I am reverting back to the old phase itself (where instead of `division` we would be doing `multiplication`). I will open an issue for this bug and leave the code and tests unchanged in this issue.
* I still haven't done the type hinting completely and hence didn't include it in `tox`.
* Still have to replace `asserts` with `Exceptions`.
* `circleci` was annoying with the installation of `mypy-extensions`, and hence I have flooded `tox.ini` with the installation of `mypy-extensions`. It's working for now, but might not be suboptimal.
* Last but not least, we need a mechanism to validate the `constants` of the curves. For example, we need to be sure that the `field_modulus` is a prime number. I feel that these shouldn't be part of the `tests` and maybe we have to test the `validity` of these constants everytime we import the whole module or it's submodule. I was thinking maybe we can validate these constants in `__init__.py`. What do you say?

Bhargavasomu:
`Type Hinting` is done now.

Bhargavasomu:
@pipermerriam I have addressed all the above mentioned issues (except the `bug` of `multiplication` instead of `division`). Please take a look and we can maybe wrap this up.

vbuterin:
Responded to the issue in the other thread; basically, the code is correct and the comment is wrong, and it needs to use `* w**2` in the BLS curve and `/ w**2` in the bn128 curve is because the relationship between `b` and `b2` in the two curves is different.

pipermerriam:
> Instead of using `classmethod` or `staticmethod` decorators, I was thinking that maybe we could instantiate an object in `_init__.py` like `bn128 = BN128_Curve()`, and the users could use this object itself directly. Need your suggestion on this.

These two things are not mutually exclusive.  Even in the `bn128 = BN128_Curve()` model, I think the code benefits from marking the methods as `staticmethod/classmethod` appropriately if they don't use any `self` properties.

pipermerriam:
> Last but not least, we need a mechanism to validate the constants of the curves. For example, we need to be sure that the `field_modulus` is a prime number. I feel that these shouldn't be part of the tests and maybe we have to test the validity of these constants everytime we import the whole module or it's submodule. I was thinking maybe we can validate these constants in `__init__.py`. What do you say?

Whatever the solution, it sounds like something that cannot be a runtime check.  If it is *fast* an import time check seems fine.  Having a test which checks this also seems fine.

Bhargavasomu:
@pipermerriam yes I have understood that, and I have added `staticmethod` and `classmethod` decorators as needed. All the issues that `I` have mentioned in the above comment have been resolved. Need you to review it. Thankyou!

Bhargavasomu:
@pipermerriam the validation checks that we are doing are fast enough for an `import`, and once the module is imported, it won't be called multiple times. But just to be sure I have `cached` that validation function.

py_ecc/bn128_curve.py
```diff
@@ -0,0 +1,180 @@
...
+    G1 = curve_properties[curve_name]["G1"]
+    G2 = curve_properties[curve_name]["G2"]
+    Z1 = curve_properties[curve_name]["Z1"]
+    Z2 = curve_properties[curve_name]["Z2"]
+    ate_loop_count = curve_properties[curve_name]["ate_loop_count"]
+    log_ate_loop_count = curve_properties[curve_name]["log_ate_loop_count"]
+    pseudo_binary_encoding = curve_properties[curve_name]["pseudo_binary_encoding"]
+
+    def __init__(self) -> None:
+        super().__init__()
```
pipermerriam: This looks like it can be removed since it is a direct call to `super()` and thus doesn't change the base implementation.

py_ecc/bn128_curve.py
```diff
@@ -0,0 +1,180 @@
...
+from py_ecc.typing import (
+    Field,
+    Optimized_Point3D,
+    Point2D,
+)
+
+
+class BN128_Curve(BaseCurve):
+    curve_name = "bn128"
+    curve_order = curve_properties[curve_name]["curve_order"]
```
pipermerriam: I am finding this structure awkward and somewhat hard to read.  It seems like this would be significantly simpler as something like.

```python
# py_ecc/field_properties.py
bn128_props = typing.NamedTuple(  # using 3.5 syntax because I think we have to support it
    'bn128_props',
    (
        ('b', int),
        ...
    )
)

# py_ecc/bn128_curve.py
from py_ecc.field_properties import bn128_props

class BN128_Curve(BaseCurve):
    field_modulus = bn128_props.b
    ...
```

Benefits I see:

- Removes the string based field lookups: `thing.prop` instead of `thing["prop"]`
- Removes the nested lookups: `thing.prop` instead of `outer_things[thing_name]["prop"]`

py_ecc/field_elements.py
```diff
@@ -64,33 +57,39 @@ def __radd__(self, other: IntOrFQ) -> "FQ":
...
+            self.n * prime_field_inv(on, self.field_modulus) % self.field_modulus,
+            self.curve_name
+        )
 
     def __truediv__(self, other: IntOrFQ) -> "FQ":
         return self.__div__(other)
 
     def __rdiv__(self, other: IntOrFQ) -> "FQ":
         on = other.n if isinstance(other, FQ) else other
         assert isinstance(on, int), on
```
pipermerriam: still some leftover runtime assertions that should be converted to explicit exception raising.

py_ecc/field_elements.py
```diff
@@ -40,21 +23,31 @@ def inv(a: int, n: int) -> int:
...
+    field_modulus = None
+    curve_name = None
+
+    def __init__(self, val: IntOrFQ, curve_name: str) -> None:
+        """
+        curve_name can be either 'bn128' or 'bls12_381'
+        This is needed to obtain field_modulus, FQ2_MODULUS_COEFFS
+        and FQ12_MODULUS_COEFFS from the curve properties
+        """
+        self.curve_name = curve_name
```
pipermerriam: My *gut* is telling me that this `curve_name` based approach could be improved.  The primary downside that I want to get rid of is how much we seem to have to pass this parameter around.  If this was instead a property of the class....

```python
class FQ(ABC):
    field_modulus = None

    def __init__(self, val):
        if self.field_modulus is None:
           raise SomeError("...")
```

This sets a simple baseline for ensuring that `FQ` classes and their derivatives don't get instantiated without their `field_modulus`

```python
class FQ_BN128(FQ):
    field_modulus = field_properties["bn128"]["field_modulus"]
```

Now we have an `FQ_BN128` which represents the `FQ` for the bn128 curve.

```python
class FQ:
    ...

    def __add__(self, other):
        on = other.n if isinstance(other, FQ) else other
        return type(self)(self.n + on) % self.field_modulus)

    @classmethod
    def one(cls):
        return cls(1)
```

This allows us to remove the passing around of the `self.curve_name` within the `FQ` object methods quite easily.

I did a quick search and this seems to cleanly remove the need to pass the curve name around.  What do you think?

py_ecc/validate_constants.py
```diff
@@ -0,0 +1,135 @@
...
+from py_ecc.field_properties import (
+    field_properties,
+)
+
+
+def validate_field_properties() -> None:
+    for curve_name in field_properties:
+        # Check if field_modulus is prime
+        field_modulus = field_properties[curve_name]["field_modulus"]
+        if pow(2, field_modulus, field_modulus) != 2:
```
pipermerriam: I feel dense.  How does this check for *primeness*?

setup.py
```diff
@@ -4,15 +4,18 @@
 
 extras_require = {
     'test': [
-        "pytest==3.2.2",
-        "pytest-xdist"
+        "mypy-extensions>=0.4.1",
```
pipermerriam: This will have to go in the core `install_requires`, otherwise the import of the `TypedDict` type will fail on normal installation.

py_ecc/validate_constants.py
Bhargavasomu: @pipermerriam I believe that is the `Fermat's Little Theorem`.

py_ecc/field_elements.py
Bhargavasomu: @pipermerriam most of these asserts here are only minimal, and should infact be replaced by `raise` exceptions as part of other functions like `__add__`, `__mul__` etc. Or we ignore them completely all together. I would say that we add those `Exceptions`, as it is an important check. But how about we keep that for another PR or issue? (Which I would be taking up) This is because once we get the basic structure clean, rest all can be done easily.

py_ecc/field_elements.py
Bhargavasomu: @pipermerriam I agree with this, but I was left with no other option. I have thought of this approach, but it would involve a lot of classes in the code. One class should be declared for each of the curves `bn128`, `bls12_381`, `optimized_bn128`, `optimized_bls12_381` curve ; that too for only `FQ`. For `FQP`, `FQ2` and `FQ12`, the same process needs to be repeated, and we would be ending up with `16 classes` in total. Hence I just resented to this approach.

I know this design of using `curve_name` leads to a lot parameter passing, but I felt it was the best choice, given the above reasons. Please correct me if wrong.

py_ecc/validate_constants.py
carver: If I'm reading this correctly, the test correctly detects **some, but not all** cases where `field_modulus` is not prime. Maybe the comment could be updated since this is kind of a "smoke test" rather than a proof of primeness.

(For example, see `341` which is not prime, but would not raise this exception)

Bhargavasomu: @carver nice catch. Didn't realize that. Will fallback to the traditional `O(sqrt(n))` way of checking a prime number. Thankyou!

Bhargavasomu: Since the `field_modulus` and the other numbers that we are checking for primes, are quite large, the `O(sqrt(n))` method is taking a lot of time. We might as well stay with the present `smoke test`.

py_ecc/field_elements.py
Bhargavasomu: @pipermerriam done!

Bhargavasomu:
@pipermerriam I have modified the code based on the above review. Please check again now. Thankyou!

Bhargavasomu:
`Assertions` have now been replaced with suitable `Exceptions`, and additional `Exceptions` have been added wherever necessary. Please have a look.

Bhargavasomu:
@pipermerriam sorry for the multiple pings, but the `WIP` status has been removed, and is read for hopefully a final round of review. Thankyou.

Bhargavasomu:
@pipermerriam any update on this?

pipermerriam:
Taking a look now

pipermerriam:
I'm unsure how to proceed right now.  This codebase implements cryptographic primitives and thus is extremely sensitive and **needs** to be correct.

This pull request basically touches *everything* and does so in such a way that it's non-trivial to compare the before/after implementations to ensure that none of the functionality has changed.  Similarly with the constants, there are lots of very large integer numbers and it's non-trivial to confirm that none of those numbers have changed.

Also, I recall the performance of this version being significantly worse than the current implementation.  Has that been addressed?

Maybe there is a way to do this in a more iterative manner that touches a smaller chunks at a time...   As it stands I can't justify merging this and I don't have an immediate suggestion for how to proceed.

Bhargavasomu:
@pipermerriam the `performance issue` has been resolved. Regarding the `Iterative Procedure`, since we have different curves, and we want to do it in a generic way, we would need to change all of them once I guess. 

> Similarly with the constants, there are lots of very large integer numbers and it's non-trivial to confirm that none of those numbers have changed.

We have a smoke test in place so as to ensure that the numbers are primes (with a general wave of hand though.) Further I have mimicked all the tests wrt to the new refactored way. But anyways I am ready to redo the work in a Iterative manner if you propose it.

carver:
> I am ready to redo the work in a Iterative manner if you propose it.

I agree that a series of smaller PRs would be the best way to get high quality code reviews.

I roughly aim to limit to 100's of lines of code changed.

There are times when exceptions are reasonable, and we can reasonably review larger PRs. One example is when there is *clearly* no change in functionality, like:
- moving files from one module to another
- squashing a bunch of flake8 warnings after updating the config

Those large PRs should make exactly that one change, though.

carver:
One option would be to leave the old curves implementations in, and add the new curve implementations one at a time. (yes, I know the first PR will still be big, but hopefully still significantly smaller)  After the new ones are all added in each of their PRs, we can delete the old ones.

But this is just a guess. It will likely take some digging and playing on your part to figure out how to get it into smaller chunks. One thing I want to be able to easily do is verify that a single new API works the same way as the old API. Ideally, that would mean roughly one new API added (xor the refactor of one component) in each PR.

---

Also, it looks like a few basics can be split out, like the tox.ini or setup.py changes. Feel free to ignore whole files in flake8 if we know they will be eventually deleted (like some of the `tests/`).

---

If you really can't find any way to slice it up, then we will get to it eventually, but I can't make any promises about when. It's too sensitive to do a "merge and fix-forward" kind of approach.

---

Another approach to consider: split any API changes from major refactors. A major refactor shouldn't require any test changes, if the tests are written well.

Adding new APIs as an augmentation to old APIs can happen in their own PR, and you can write tests that compare the new and old APIs.

Then eventually the old APIs can be deleted at a major version bump. This is a common approach we take for other libraries.

PR #38, opened 2018-12-11
Bring Python3.5 back

hwwhww:
### What was wrong?
We need Python3.5 for supporting other packages before a series of upgrades.

### How was it fixed?
Bring Python 3.5 back from ethereum/py_ecc#22.

#### Cute Animal Picture
🙀


carver:
> We need Python3.5 for supporting other packages before a series of upgrades.

If you happen to have this list, it would be handy to have it shared so that we can fix each of the packages in dependency order.

PR #39, opened 2018-12-11
Modernize release process

pipermerriam:
### What was wrong?

Old manual (error prone) release process

### How was it fixed?

Modernized to use bumpversion

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ extras_require = {
     ],
     'dev': [
         "bumpversion>=0.5.3,<1",
-        'twine',
+        "twine",
     ],
 }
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -11,10 +11,15 @@
         "flake8==3.4.1",
         "mypy==0.641",
     ],
+    'dev': [
+        "bumpversion>=0.5.3,<1",
+        'twine',
```
carver:
```suggestion
        "twine",
```

_nitpick_ the single quotes really stand out next to the bumpversion's double quotes.

pipermerriam: copy/pasta

PR #40, opened 2018-12-12
Add __mod__ to FQ and FQP for type hinting

hwwhww:
### What was wrong?
For type hinting, it requires `__mod__` to return the correct 

### How was it fixed?
Add `__mod__` for `FQP` for type hinting.

#### Cute Animal Picture

Also, on Py-EVM side, https://github.com/hwwhww/py-evm/commit/a326a0e90c9dddad184b0dd05e018bf40df9a719 would work with this PR.


pipermerriam:
I assume you want a patch release for this asap?

PR #41, opened 2019-01-13
Remove Redundancy of Fields among curves

Bhargavasomu:
### What was wrong?
As part of the refactoring, the first step is to remove the redundant `field_elements.py` and `optimized_field_elements.py` files and make it possible to develop an API which any curve can use, instead of mere copy pasting with few properties changes.

### How was it fixed?
The `field_properties` has been generated. Also the `field_elements` and the `optimized_field_elements` files generate the fields `FQ` or `FQP` by making use of the `curve name`. Further the tests corresponding to these have been created in another file named `test_field_elements.py` which has been taken out from `test_bn128_and_bls12_381.py` (It shouldn't be hard to corelate the tests, just changed the syntax which is caused due to the breaking API change).

### Note
This API hasn't still been integrated into the curves logic. I have just made the API change and mimiced the corresponding changes to the test cases. If this API is ok, then I will integrate them with the other curves. Please let me know the **exhaustive** list of references or explanations if you cannot co-relate anything to the already existing codebase. I will give you an explanation for all the questions posed.

#### Cute Animal Picture


Bhargavasomu:
@pipermerriam @carver review please

carver:
First thought (I'll come back and review more, soon):

I think you might be able to fit the new curves into the old tests by modifying this area:
https://github.com/ethereum/py_ecc/blob/5aba1a732a85c8e318874d6789000f0d07174fbf/tests/test_bn128_and_bls12_381.py#L8-L15

To do something like:

```py
from functools import partial

@pytest.fixture(params=[
  bn128, optimized_bn128, bls12_381, optimized_bls12_381,
  "bn128", "optimized_bn128"])
def curve(request):
    return request.param

@pytest.fixture
def FQ(curve):
  if isinstance(curve, str):
    if curve.startswith('optimized_'):
      return partial(optimized_field_elements.FQ, curve_name=curve[10:])
    else:
      return partial(field_elements.FQ, curve_name=curve)
  else:
    return lib.FQ
```

That would allow testing the new models and the old models side by side, using the same tests, which is ideal during a refactor.

BTW, note that the partial only currently works with a keyword argument. It would work positionally if the argument order of the init were switched. That would be a very reasonable thing to do.

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+class FQ12(FQP):
+    """
+    The 12th-degree extension field
+    """
+    degree = 12
+
+    def __init__(self, coeffs: Sequence[IntOrFQ], curve_name: str) -> None:
+        FQ12_MODULUS_COEFFS = field_properties[curve_name]["fq12_modulus_coeffs"]
```
carver: I don't love that this would just through a `KeyError` if someone used the wrong `curve_name`. I think we'll have that problem in a number of places.

Let's think about alternatives to this setup. Maybe something like:

```py
class Curve(NamedTuple):
  field_modulus: int
  fq12_modulus_coeffs: Tuple[int, ...]
  fq_modulus_coeffs: Tuple[int, int]

bls12_381 = Curve(
  field_modulus = 21888242871839275222246405745257275088696311157297823662689037894645226208583,
  fq2_modulus_coeffs = (1, 0),
  fq12_modulus_coeffs = (2, 0, 0, 0, 0, 0, -2, 0, 0, 0, 0, 0),  # Implied + [1]
)
```

Which makes this method something like:
```py
def __init__(self, coeffs: Sequence[IntOrFQ], curve: Curve) -> None:
  super().__init__(coeffs, curve, curve.fq12_modulus_coeffs)
```
Of course, it comes at the downside of having to import the curve to create an `FQ`. But for the caller, it means fewer magic strings and less likelihood for a typo with a confusing error. For the library, hopefully better readability (at least this method got simpler).

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+    field_properties,
+)
+
+from py_ecc.utils import (
+    deg,
+    prime_field_inv,
+)
+
+
+IntOrFQ = Union[int, "FQ"]
```
carver: For a future PR: we can pull this into a `py_ecc.types` module

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+class FQ(object):
```
carver: Looks like we can just delete this `FQ` and import it from the regular (unoptimized) implementation.

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,355 @@
...
+                 curve_name: str,
+                 modulus_coeffs: Sequence[IntOrFQ]=None) -> None:
+        if len(coeffs) != len(modulus_coeffs):
+            raise Exception(
+                "coeffs and modulus_coeffs aren't of the same length"
+            )
+        self.coeffs = tuple(FQ(c, curve_name) for c in coeffs)
+        self.curve_name = curve_name
+        # The coefficients of the modulus, without the leading [1]
+        self.modulus_coeffs = modulus_coeffs
```
carver: Could be good to wrap this in a `tuple()` like the optimized one.

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+        and FQ12_MODULUS_COEFFS from the curve properties
+        """
+        self.curve_name = curve_name
+        self.field_modulus = field_properties[curve_name]["field_modulus"]
+
+        if len(coeffs) != len(modulus_coeffs):
+            raise Exception(
+                "coeffs and modulus_coeffs aren't of the same length"
+            )
+        self.coeffs = tuple(coeffs)
```
carver: Maybe worth commenting here that we're explicitly not converting coefficients to `FQ()` here, for performance reasons.

... maybe even explicitly cast to `int`.

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+                 coeffs: Sequence[IntOrFQ],
+                 curve_name: str,
+                 modulus_coeffs: Sequence[IntOrFQ]=None) -> None:
+        """
+        curve_name can be either 'bn128' or 'bls12_381'
+        This is needed to obtain field_modulus, FQ2_MODULUS_COEFFS
+        and FQ12_MODULUS_COEFFS from the curve properties
+        """
+        self.curve_name = curve_name
+        self.field_modulus = field_properties[curve_name]["field_modulus"]
```
carver: Again, I quite appreciate it that you kept it as similar as possible to the old `FQP`, for the first review.

*suuper nitpicky* but I'd be in favor of reordering the `init` lines so that they match up with the `FQP` in the non-optimized version (at least, where the lines are the same).

If you don't feel like it, no worries, we can add an issue and someone can come back later to clean it up (maybe a `Good First Issue`).

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+            )
+
+        return type(self)([
+            int(x - y) % self.field_modulus
+            for x, y
+            in zip(self.coeffs, other.coeffs)
+        ], self.curve_name)
+
+    def __mod__(self, other: Union[int, "FQP"]) -> "FQP":
+        return self.__mod__(other)
```
carver: What the heck is this? :eyes: :fire:  It just recurses until it explodes on stack depth.

I mean, I see it's not new code: it was in the old implementation. It just doesn't make sense, and clearly must not be getting tested.

Maybe we should just raise a `NotImplementedError` until we can figure out what should be happening here. (Note that there are a couple of these)

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+                for j, elj in inner_enumerate:
+                    b[i + j] += int(eli * elj)
+            # MID = len(self.coeffs) // 2
+            for exp in range(self.degree - 2, -1, -1):
+                top = b.pop()
+                for i, c in self.mc_tuples:
+                    b[exp + i] -= top * c
+            return type(self)([x % self.field_modulus for x in b], self.curve_name)
+        else:
+            raise ValueError(
```
carver: *nit* `TypeError`

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+        if isinstance(other, int):
+            return type(self)([
+                int(c) * prime_field_inv(other, self.field_modulus) % self.field_modulus
+                for c
+                in self.coeffs
+            ], self.curve_name)
+        elif isinstance(other, type(self)):
+            return self * other.inv()
+        else:
+            raise ValueError(
```
carver: *nit* `TypeError`

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+        return [x % self.field_modulus for x in o[:deg(o) + 1]]
+
+    # Extended euclidean algorithm used to find the modular inverse
+    def inv(self) -> "FQP":
+        lm, hm = [1] + [0] * self.degree, [0] * (self.degree + 1)
+        low, high = (
+            cast(List[IntOrFQ], list(self.coeffs + (0,))),
+            cast(List[IntOrFQ], list(self.modulus_coeffs + (1,))),
+        )
+        low, high = list(self.coeffs + (0,)), self.modulus_coeffs + (1,)  # type: ignore
```
carver: This line appears to be redundant, can be deleted. (I see it was in the original)

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+            new = [int(x) % self.field_modulus for x in new]
+            lm, low, hm, high = nm, new, lm, low
+        return type(self)(lm[:self.degree], self.curve_name) / low[0]
+
+    def __repr__(self) -> str:
+        return repr(self.coeffs)
+
+    def __eq__(self, other: "FQP") -> bool:     # type: ignore # https://github.com/python/mypy/issues/2783 # noqa: E501
+        if not isinstance(other, type(self)):
+            raise ValueError(
```
carver: *nit* `TypeError`

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+                "Expected an FQP object, but got object of type {}"
+                .format(type(other))
+            )
+
+        for c1, c2 in zip(self.coeffs, other.coeffs):
```
carver: I think this had a bug in the original source, but seems out of scope. Opened #45

py_ecc/fields/optimized_field_elements.py
```diff
@@ -0,0 +1,395 @@
...
+        o = type(self)([1] + [0] * (self.degree - 1), self.curve_name)
+        t = self
+        while other > 0:
+            if other & 1:
+                o = o * t
+            other >>= 1
+            t = t * t
+        return o
+
+    def optimized_poly_rounded_div(self,
```
carver: :+1: makes total sense to move this into the class

py_ecc/fields/optimized_field_elements.py
Bhargavasomu: @carver I have observed this prior, but have just copy/pasted because to maintain symmetry between the `optimized` and the `non-optimized` version in terms of `type hinting`, `imports` etc at this point of time.

How about I change this after we complete this whole refactor thing?

Bhargavasomu:
@carver instead of using `partial`, I have manually created the classes to all the `Fields` of the curves, (thank guido that there is a `type` method :P). The code is now visually better to read and now visually a feast to those referring to the codebase. @pipermerriam I think you would be happy with this change, as I remember you asking to change this earlier as part of the other refactoring PR.

Further, I have removed the mypy checks, as they are messy right now. I will later integrate mypy, after this whole refactor thing is done with.

py_ecc/fields/field_elements.py
Bhargavasomu: @carver is it ok, if I take this up in a future PR?

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,366 @@
...
+    def __radd__(self, other: IntOrFQ) -> "FQ":
+        return self + other
+
+    def __rsub__(self, other: IntOrFQ) -> "FQ":
+        if isinstance(other, FQ):
+            on = other.n
+        elif isinstance(other, int):
+            on = other
+        else:
+            raise TypeError(
```
pipermerriam: This sequence of `if/elif/else` is repeated multiple times.  What do you think about.

```python
def normalize_point(value: IntOrFQ) -> int:
    if isinstance(value, int):
        return int
    elif isinstance(value, FQ):
        return int
    else:
        raise ...
```

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,366 @@
...
+        )
+        while deg(low):
+            r = cast(List[IntOrFQ], list(poly_rounded_div(high, low)))
+            r += [0] * (self.degree + 1 - len(r))
+            nm = [x for x in hm]
+            new = [x for x in high]
+            if len(set(
+                [len(lm), len(hm), len(low), len(high), len(nm), len(new), self.degree + 1]
+            )) != 1:
+                raise Exception("Mismatch between the lengths of lm, hm, low, high, nm, new")
```
pipermerriam: I know it is more work but it would be ideal to actually include which one was the mismatch.

py_ecc/utils.py
```diff
@@ -0,0 +1,54 @@
...
+                     b: Sequence[IntOrFQ]) -> Tuple[IntOrFQ]:
+    dega = deg(a)
+    degb = deg(b)
+    temp = [x for x in a]
+    o = [0 for x in a]
+    for i in range(dega - degb, -1, -1):
+        o[i] += int(temp[degb + i] / b[degb])
+        for c in range(degb + 1):
+            temp[c + i] -= o[c]
+    return cast(Tuple[IntOrFQ], tuple(o[:deg(o) + 1]))
```
pipermerriam: I assume this type is supposed to be `Tuple[IntOrFQ, ...]` or is this really a length-1 tuple?

Bhargavasomu: @pipermerriam this actually should be `Tuple[IntOrFQ, ...]`, but as I mentioned, is it ok to skip the `type hinting` as part of this PR?

py_ecc/fields/field_elements.py
Bhargavasomu: @pipermerriam there are many places where `if/elif/else` pattern is being used. But out of them, I feel that very few come under the pattern that you are suggesting (i.e. where the usage of `normalize_point` is apt).

But if you feel that this would make the code more readable, please let me know. Will make the changes.

py_ecc/fields/field_elements.py
Bhargavasomu: On this.

py_ecc/fields/field_elements.py
pipermerriam: - https://github.com/ethereum/py_ecc/pull/41/files/9c80ebd6ea0b461d978565b4492cd2c4f40ba5a6#diff-14f5ca34559da438b521e765e0cb1fdaR52
- https://github.com/ethereum/py_ecc/pull/41/files/9c80ebd6ea0b461d978565b4492cd2c4f40ba5a6#diff-14f5ca34559da438b521e765e0cb1fdaR65
- https://github.com/ethereum/py_ecc/pull/41/files/9c80ebd6ea0b461d978565b4492cd2c4f40ba5a6#diff-14f5ca34559da438b521e765e0cb1fdaR84

Maybe my intent wasn't clear.  The `if/elif/else` block in all of these three locations could be replaced by:

```
def __rsub(self, other: IntOrFQ) -> "FQ":
    value = normalize_point(other)
    return type(self)((value - self.n) % self.field_modulus)
```

It makes for much more concise/easy-to-read methods IMHO.

py_ecc/fields/field_elements.py
Bhargavasomu: Also please note that, we could display an error if only one of them all, don't match with the rest. Assuming that would be the scenario, I could cook something up.

Bhargavasomu:
@pipermerriam made the changes, please review again. Thankyou.

Bhargavasomu:
@carver could you please review with the new changes made? Thankyou.

Bhargavasomu:
When it comes to the `proficiency`, this `refactored code` is taking 70 seconds longer than the code on `master` to run all the tests using `pytest`. I did a quick profiling and observed that, this increase is because of the following reasons.
* We are trying to normalize the given input at many places in the `FQ operations`. For example, to add an `FQ` point with `another number (or) another FQ point`, we are first checking the type with an `isinstance` check. This is causing a lot of overhead as the whole library boils down to usage of `FQ` operations (IIUC).
* Introduction of `Error Handling` in various functions, which is further done by introducing a lot of `isinstance` checks.

@ChihChengLiang @carver could you please review this?
/cc @pipermerriam

Bhargavasomu:
@pipermerriam @carver @ChihChengLiang the main devil causing more time here was, `making the function call **normalize_FQ_point**`. This is a `costly operation` because those functions are called a lot of times which makes it `time consuming`. On removing that function and pasting the 
`actual function code` wherever needed, made the code way faster. The time taken to run the `tests` by `pytest` is now `nearly the same or faster than master` now. Please verify this and review if possible. Thankyou.

ChihChengLiang:
I know it might be too late to say and you might have already tried, but not sure if you like this idea:

If we want field elements to be a common API, in the `py_ecc/fields` they shouldn't know anything about the curve. So the proposed elements API should look like:

```python
# py_ecc/fields/field_elements.py


# Can use type variable in base class methods
TFQ = TypeVar("TFQ", bound="BaseFQ")


class BaseFQ(ABC):

    @property
    @abstractmethod
    def field_modulus(self):
        pass


class BaseFQP(ABC):
    @property
    @abstractmethod
    def degree(self):
        pass


class FQ2(FQP):
    degree = 2
    

class FQ12(FQP):
    degree = 12

```

And in every curve, we keep `some_curve_field_elements.py` and let it inherit base field elements.

```python
# py_ecc/bls12_381/bls12_381_field_elements.py
from .constants import field_modulus
from py_ecc.fields.field_elements import BaseFQ

class FQ(BaseFQ):
    field_modulus = field_modulus
```

Bhargavasomu:
@ChihChengLiang thankyou for reviewing this. I see what you are saying about `Fields` should not know anything about the curve. I could just remove the need to pass the `curve_name` in the fields, which is pretty useless.

Another major point, that I wanted to raise is that, I am minorly :-1: on keeping the `<some_curve>_field_elements.py`. I want all of those to be as part of the `Fields` API. That is why you can find the fields of all the curves in `__init__.py` of the `Fields Module` in this `PR`.

One more thing about the `Base*` convention is that, I don't think that we need a `Base*` class here. This is because, even if we were to include a `Base` class, it would have a lot of functionality and not just the `abstract functions`, which is kind of adverse to the definition of `Base Classes` IIUC (Please correct me if wrong).

py_ecc/fields/__init__.py
```diff
@@ -0,0 +1,160 @@
...
+    FQ as optimized_FQ,
+    FQP as optimized_FQP,
+    FQ2 as optimized_FQ2,
+    FQ12 as optimized_FQ12,
+)
+
+
+# Create seperate classes for all Fields for each curve
+
+bn128_FQ = type(
```
pipermerriam: Is there a reason we're using `type` here instead of normal class declaration?

```python
class bn128_FQ(FQ):
    curve_name = "bn128"
    field_modulus: field_properties["bn128"]["field_modulus"]
```

If there isn't, these should be converted to normal classes.

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+class FQ(object):
+    """
+    A class for field elements in FQ. Wrap a number in this class,
+    and it becomes a field element.
+    """
+    n = None  # type: int
+    field_modulus = None
+    # curve_name can be either 'bn128' or 'bls12_381'
```
pipermerriam: What do you think about using an `enum.Enum` for the curve names.  This would give us much stronger type safety and prevent certain types of errors like misspelling the dictionary key.

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,377 @@
...
+        else:
+            raise TypeError(
+                "Expected an int or FQ object, but got object of type {}"
+                .format(type(val))
+            )
+
+    def __add__(self, other: IntOrFQ) -> "FQ":
+        if isinstance(other, FQ):
+            on = other.n
+        elif isinstance(other, int):
```
pipermerriam: I don't know if this should be experimented with before merging this PR or something we try after but... what if we just made `FQ` a subclass of `int`.  My intuition says it could solve a whole swath of these various type checks that occur since we could treat `FQ` instances as integers.

What isn't clear to me is how difficult or necessary it would be to ensure that an `FQ` object is what is return returned from operations like `a * b` and `b * a` where one of `a, b` is an `FQ` and one is an `int`.  Stated differently, it seems potentially problematic if it is difficult to reason about whether an object we are interacting with is an integer or an `FQ` as well as whether the return value of various mathmatical operations will be an `int` or `FQ`

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,377 @@
...
+
+    def __repr__(self) -> str:
+        return repr(self.n)
+
+    def __int__(self) -> int:
+        return self.n
+
+    @classmethod
+    def one(cls) -> "FQ":
+        return cls(1)
```
pipermerriam: Not for this PR but a possible performance gain might be available if we interned (cached) these return values so that we don't need to re-instantiate the `1` and `0` values each time.

py_ecc/fields/field_elements.py
```diff
@@ -0,0 +1,377 @@
...
+        return repr(self.coeffs)
+
+    def __eq__(self, other: "FQP") -> bool:     # type: ignore # https://github.com/python/mypy/issues/2783 # noqa: E501
+        if not isinstance(other, type(self)):
+            raise TypeError(
+                "Expected an FQP object, but got object of type {}"
+                .format(type(other))
+            )
+
+        for c1, c2 in zip(self.coeffs, other.coeffs):
```
pipermerriam: I *think* that since we have the `isinstance` check up top we're safe here, but worth pointing out that `zip` discards extra values if one of the values is longer so unless we know that these are the same length there's an error edge cases here.

py_ecc/fields/field_elements.py
Bhargavasomu: The `curve_name` attribute has been removed from the API. This is because the `fields` should not be knowing about the `curve` which is going to use the fields. Hence it has been removed which makes this passing around of `curve_name` obsolete.

py_ecc/fields/field_elements.py
Bhargavasomu: @pipermerriam https://github.com/ethereum/py_ecc/issues/45 should take care of this. (Probably a separate PR for this.)

py_ecc/fields/field_elements.py
Bhargavasomu: @pipermerriam I see what you are saying. It should work as per theory. But I would like this experimentation to be done after the refactoring. Opened https://github.com/ethereum/py_ecc/issues/54 for this.

py_ecc/fields/field_elements.py
Bhargavasomu: Opened #55 for this

Bhargavasomu:
@pipermerriam this should be ready to merge.

PR #42, opened 2019-01-13
Fix misconfigured checks for point at infinity

wemeetagain:
### What was wrong?
The unoptimized elliptic curve point at infinity is represented by None.
The optimized elliptic curve point at infinity is represented by (x, y, 0)

There are a few functions in the optimized curve library that incorrectly check for infinity with a null check.
### How was it fixed?
The null checks are replaced with `is_inf` checks.

#### Cute Animal Picture

### DIFF
--- a/py_ecc/optimized_bn128/optimized_curve.py
+++ b/py_ecc/optimized_bn128/optimized_curve.py
@@ -142,15 +142,15 @@ w = FQ12([0, 1] + [0] * 10)
 
 # Convert P => -P
 def neg(pt: Optimized_Point3D[Optimized_Field]) -> Optimized_Point3D[Optimized_Field]:
-    if pt is None:
-        return None
+    if is_inf(pt):
+        return pt
     x, y, z = pt
     return (x, -y, z)
 
 
 def twist(pt: Optimized_Point3D[FQP]) -> Optimized_Point3D[FQP]:
-    if pt is None:
-        return None
+    if is_inf(pt):
+        return pt
     _x, _y, _z = pt
     # Field isomorphism from Z[p] / x**2 to Z[p] / x**2 - 18*x + 82
     xcoeffs = [_x.coeffs[0] - _x.coeffs[1] * 9, _x.coeffs[1]]
--- a/py_ecc/optimized_bn128/optimized_curve.py
+++ b/py_ecc/optimized_bn128/optimized_curve.py
@@ -149,8 +149,6 @@ def neg(pt: Optimized_Point3D[Optimized_Field]) -> Optimized_Point3D[Optimized_F
 
 
 def twist(pt: Optimized_Point3D[FQP]) -> Optimized_Point3D[FQP]:
-    if is_inf(pt):
-        return pt
     _x, _y, _z = pt
     # Field isomorphism from Z[p] / x**2 to Z[p] / x**2 - 18*x + 82
     xcoeffs = [_x.coeffs[0] - _x.coeffs[1] * 9, _x.coeffs[1]]
--- a/py_ecc/optimized_bn128/optimized_curve.py
+++ b/py_ecc/optimized_bn128/optimized_curve.py
@@ -142,8 +142,6 @@ w = FQ12([0, 1] + [0] * 10)
 
 # Convert P => -P
 def neg(pt: Optimized_Point3D[Optimized_Field]) -> Optimized_Point3D[Optimized_Field]:
-    if is_inf(pt):
-        return pt
     x, y, z = pt
     return (x, -y, z)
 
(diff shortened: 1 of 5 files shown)


carver:
Are you up for adding some test cases that would fail these 4 cases?

carver:
Pardon my push to your branch, feel free to revert it if I have something wrong. I noticed that the `neg()` changes weren't causing the old test suite to turn red, so I dug a bit into why. My interpretation: the `neg()` change isn't fixing a logical bug, but is an optimization. So I added in a test for the optimization, and a note about why it was there. Now it turns the old implementation red and the new implementation is green.

Unfortunately, I don't have enough expertise to confirm the rest of the changes, so I'll have to lean on someone else to take a look for the final :+1:

carver:
Let's see if we can avoid bothering vbuterin for it. Maybe @ChihChengLiang can verify the correctness of the changes?

wemeetagain:
Yeah, I'd agree with your assessment on changes to `neg`, it's more of an optimization/consistency thing than anything else. The point at infinity seems to be any (x, y, 0), (according to `is_inf`) so it shouldn't(?) really matter if its (x, y, 0) or (x, -y, 0)

The change to unoptimized `double` function is a legit bug. The function assumes a tuple, where a point at infinity is None.

There does appear to be a problem with what's been changed in `twist`.
In the non `is_inf` case, we're outputing a coordinate where x,y,and z are in Fq12. We should still be doing this in the `is_inf` case, rather than blindly returning the point as is.
(There may be an opportunity for tightening of the function's type hints to further elucidate this.)

As someone without a deep understanding of the subject matter (learning, but slowly), but with more of a background and intuition for software in general, this seems like a risky addition. These additions are at best, tiny optimizations, and at worst, serious blunders.
Perhaps we can save a few of the tests regardless, the unoptimized `double` is_inf check, but we definitely need more eyes on this before merging.

wemeetagain:
I removed the `is_inf` conditional from `twist`, based on my comment above. Fixed tests to match.

Also added a test showing the root of this PR's 'fix', showing that None points should correctly throw an exception in the optimized curve functions.

pipermerriam:
cc @Bhargavasomu as you may be able to review the correctness of this.

py_ecc/optimized_bn128/optimized_curve.py
```diff
@@ -142,15 +142,13 @@ def normalize(pt: Optimized_Point3D[Optimized_Field]) -> Optimized_Point2D[Optim
...
-        return None
+    if is_inf(pt):
+        return pt
     x, y, z = pt
     return (x, -y, z)
 
 
 def twist(pt: Optimized_Point3D[FQP]) -> Optimized_Point3D[FQP]:
-    if pt is None:
-        return None
```
ChihChengLiang: This removal of None check makes sense to me. Since points at infinity Z1 and Z2 are tuples in this case, None check is pointless.

ChihChengLiang: This removal of None check makes sense to me. Since points at infinity Z1 and Z2 are tuples in this case, None check is pointless.

py_ecc/optimized_bn128/optimized_curve.py
```diff
@@ -142,15 +142,13 @@ def normalize(pt: Optimized_Point3D[Optimized_Field]) -> Optimized_Point2D[Optim
 
 # Convert P => -P
 def neg(pt: Optimized_Point3D[Optimized_Field]) -> Optimized_Point3D[Optimized_Field]:
-    if pt is None:
-        return None
+    if is_inf(pt):
+        return pt
```
ChihChengLiang: If I'm doing the profiling right, is_inf check is no faster than just do the calculation
https://gist.github.com/ChihChengLiang/ac2a89f3ff90cb0c559f1d2423f5aa1c#file-profile_inf_check-py

wemeetagain: Makes sense to just remove the check entirely, as done in `twist`

PS: thanks for showing your work RE profiling

ChihChengLiang: Sorry, the previous profiling is incorrect for not using real `Z1` and `is_inf`. I did a new one, and the conclusion still holds. no-check is faster.
https://gist.github.com/ChihChengLiang/ac2a89f3ff90cb0c559f1d2423f5aa1c#file-fixed_profile_inf_check-py

tests/test_bn128_and_bls12_381.py
```diff
@@ -160,6 +175,34 @@ def test_G12_object(G12, b12, eq, add, double, multiply, is_on_curve, is_inf, cu
...
 
 
+def test_Z1_object(add, eq, double, FQ, G1, is_inf, multiply, neg, twist, Z1):
+    assert eq(G1, add(G1, Z1))
+    assert eq(Z1, double(Z1))
+    assert eq(Z1, multiply(Z1, 0))
+    assert eq(Z1, multiply(Z1, 1))
+    assert eq(Z1, multiply(Z1, 2))
+    assert eq(Z1, multiply(Z1, 3))
+    assert is_inf(neg(Z1))
```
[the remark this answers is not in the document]
wemeetagain: From the type hints, `twist` should not be called on a FQ point, only FQP points (FQ2 and FQ12).
Also, FQ objects don't have a `coeffs` attribute, so a call with a FQ will result in an error.
Dropped for these reasons.

PR #61, opened 2019-03-07
Remove the curve_name from the fields API

Bhargavasomu:
### What was wrong?
In the `__init__.py` of the `fields API` the classes have been declared using `type`. This has to be replaced with the actual class definitions.

### How was it fixed?
Replace the `type` with `class` in the `__init__.py`.

#### Cute Animal Picture

### DIFF
--- a/py_ecc/fields/field_elements.py
+++ b/py_ecc/fields/field_elements.py
@@ -22,14 +22,8 @@ class FQ(object):
     """
     n = None  # type: int
     field_modulus = None
-    # curve_name can be either 'bn128' or 'bls12_381'
-    # This is needed to obtain field_modulus, FQ2_MODULUS_COEFFS
-    # and FQ12_MODULUS_COEFFS from the curve properties
-    curve_name = None
 
     def __init__(self, val: IntOrFQ) -> None:
-        if self.curve_name is None:
-            raise AttributeError("Curve Name hasn't been specified")
         if self.field_modulus is None:
             raise AttributeError("Field Modulus hasn't been specified")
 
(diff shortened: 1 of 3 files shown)


py_ecc/fields/field_elements.py
```diff
@@ -187,26 +181,23 @@ class FQP(object):
...
         if self.field_modulus is None:
             raise AttributeError("Field Modulus hasn't been specified")
 
         if len(coeffs) != len(modulus_coeffs):
             raise Exception(
                 "coeffs and modulus_coeffs aren't of the same length"
             )
-        # Encoding all coefficients in type FQ (in regards to the curve name too)
+        # Encoding all coefficients in the corresponding type FQ
         self.FQP_corresponding_FQ_class = type(
```
pipermerriam: Not for this PR but, This appears to be a class-level value (i.e. it is constant across all instances of this class since it only differs on the `field_modulus` property.).

If that is the case, then it *should* be possible to populate this at the class level either by doing it manually *or* by doing it using a metaclass.  Rough untested code looks something like:


```python
class FQPMetaclass(type):
    def __new__(cls, name, bases, attrs):
        field_modulus = attrs.get('field_modulus', cls.field_modulus)
        if field_modulus is not None:
            attrs['FQP_corresponding_FQ_class'] = type("FQP_corresponding_FQ_class", (FQ,), {'field_modulus": field_modulus})
        return super().__new__(name, bases, attrs)
```

Bhargavasomu: Opened #62 for this.

PR #67, opened 2019-05-06
fast-verification-of-multiple-bls-signatures

ChihChengLiang:
### What was wrong?

https://ethresear.ch/t/fast-verification-of-multiple-bls-signatures/5407

#### Cute Animal Picture


py_ecc/bls/api.py
```diff
@@ -109,3 +113,48 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
...
+        signatures: Sequence[BLSSignature],
+        pubkeys_and_messages: Tuple[Sequence[BLSPubkey], Sequence[Hash32]],
+        domain: int)-> bool:
+    if len(signatures) != len(pubkeys_and_messages):
+        raise ValidationError(
+            "len(signatures) (%s) should be equal to len(pubkeys_and_messages) (%s)" % (
+                len(signatures), len(pubkeys_and_messages)
+            )
+        )
+    try:
```
pipermerriam: This is generally an anti-pattern.  `try/except` blocks int almost all situations should be as small as possible.  Large blocks like this often result in errors like accidentally catching exceptions that you don't intend to and then handling them even though they aren't the thing you intended to handle.

py_ecc/bls/api.py
```diff
@@ -109,3 +113,48 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
...
+        random_ints = (1,) + tuple(2**randbelow(64) for _ in range(len(signatures) - 1))
+        o = FQ12.one()
+        for index, pm in enumerate(pubkeys_and_messages):
+            pubkeys, message_hashes = pm  # type: ignore
+            len_msgs = len(message_hashes)
+            if len(pubkeys) != len_msgs:
+                raise ValidationError(
```
pipermerriam: This is effectively a `GOTO` statement which could just be replaced with a `return False` since that's what the `except` block does.

py_ecc/bls/api.py
```diff
@@ -109,3 +113,48 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
...
+        raise ValidationError(
+            "len(signatures) (%s) should be equal to len(pubkeys_and_messages) (%s)" % (
+                len(signatures), len(pubkeys_and_messages)
+            )
+        )
+    try:
+        random_ints = (1,) + tuple(2**randbelow(64) for _ in range(len(signatures) - 1))
+        o = FQ12.one()
+        for index, pm in enumerate(pubkeys_and_messages):
+            pubkeys, message_hashes = pm  # type: ignore
```
pipermerriam: We've been ensuring we add a comment explaining why any `# type: ignore` statements are present so it's clear why they exist.

py_ecc/bls/api.py
```diff
@@ -109,3 +113,48 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
...
+        pubkeys_and_messages: Tuple[Sequence[BLSPubkey], Sequence[Hash32]],
+        domain: int)-> bool:
+    if len(signatures) != len(pubkeys_and_messages):
+        raise ValidationError(
+            "len(signatures) (%s) should be equal to len(pubkeys_and_messages) (%s)" % (
+                len(signatures), len(pubkeys_and_messages)
+            )
+        )
+    try:
+        random_ints = (1,) + tuple(2**randbelow(64) for _ in range(len(signatures) - 1))
```
pipermerriam: The `range` call is un-neccessary.  You could just do `... for _ in signatures[:-1]` and have the same effect.

py_ecc/bls/api.py
```diff
@@ -109,3 +113,48 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
...
+                    "len(pubkeys) (%s) should be equal to len(message_hashes) (%s)" % (
+                        len(pubkeys), len_msgs
+                    )
+                )
+            for m_pubs in set(message_hashes):
+
+                # aggregate the pubs
+                group_pub = Z1
+                for i in range(len_msgs):
+                    if message_hashes[i] == m_pubs:
```
pipermerriam: IIUC this code is finding signatures with common public keys and grouping them.  The current implementation seems to have `~O(n ** 2)` complexity since it loops over the hashes once for each hash.  You should be able to reduce this using something like:

- use `cytoolz.groupby(operator.itemgetter(1), enumerate(message_hashes))` to group the message hashes and retain their indices.
- loop over that data structure once, pulling out the indices to produce the `group_pub` for each set of common hashes.

tests/test_bls.py
```diff
@@ -238,3 +240,38 @@ def test_multi_aggregation(msg_1, msg_2, privkeys_1, privkeys_2):
...
+
+def test_multi_multi():
+    domain = 0
+    validator_indices = tuple(range(10))
+    privkeys = tuple(2**i for i in validator_indices)
+    pubkeys = [privtopub(k) for k in privkeys]
+
+    class Attestation:
+        def __init__(self, msg_1, msg_2):
+            msg_1_validators = sample(validator_indices, 3)
```
pipermerriam: Anytime we use randomness in tests we should be using the mechanisms provided by `hypothesis` since they allow us to reproduce failures when they occur.  Otherwise failures due to random data will be opaque and we won't be able to recreate the conditions that produced them.

See: https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.random_module

py_ecc/bls/api.py
```diff
@@ -109,3 +113,48 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
         return final_exponentiation == FQ12.one()
     except (ValidationError, ValueError, AssertionError):
         return False
+
+
+def verify_multiple_multiple(
```
hwwhww: Renaming ideas: `optimized_verify_multiple`, `fast_verify_multiple`...

pipermerriam: If it's faster why not just replace the existing implementation instead of maintaining two?

Also, if there's an existing one and we are keeping both around, seems prudent to test them against each other to ensure they have equal behavior (sorry if I missed the test that does this)

ChihChengLiang: The new function is an optimized version of doing multiple times of `verify_multiple`. The later verifies a single attestation and the former verifies many attestations in a block.

ChihChengLiang:
benchmark before rename and pr feedback
- profile_verify_multiple 177 sec
- profile_verify_multiple_multiple 92 sec

py_ecc/bls/api.py
```diff
@@ -80,32 +92,68 @@ def aggregate_pubkeys(pubkeys: Sequence[BLSPubkey]) -> BLSPubkey:
...
+
 def verify_multiple(pubkeys: Sequence[BLSPubkey],
                     message_hashes: Sequence[Hash32],
                     signature: BLSSignature,
                     domain: int) -> bool:
-    len_msgs = len(message_hashes)
 
-    if len(pubkeys) != len_msgs:
+    o = FQ12.one()
+    for message_hash, group_pubkeys in _group_by_messages(message_hashes, pubkeys):
```
ChihChengLiang: @hwwhww This groupby step is actually useless. The use cases in the spec suggest that `bls_verify_multiple` shouldn't group the message_hash

pipermerriam:
@ChihChengLiang glancing at the library I don't see any benchmarking tools which suggests you did the benchmarking manually.  Would you mind contributing some version of that into the core library so we can have a starting point for future library optimizations?

ChihChengLiang:
@pipermerriam manually in this PR 😂`scripts/benchmark_multi_multi.py` https://github.com/ethereum/py_ecc/pull/67/files#diff-6603fa8fb24b1bbf0f65499a385162e4

py_ecc/bls/api.py
ChihChengLiang: I do desperately want to get rid of `verify_multiple_multiple` and rename it properly. Besides the name of the function, the function parameter `pubkeys_and_messages: Sequence[Tuple[Sequence[BLSPubkey], Sequence[Hash32]]]` is also confusing. Should we create some objects like 
```python
class SignedMessage:
    signature: BLSSignature
    public_keys: Sequence[BLSPubkey]
    message_hashes: Sequence[Hash32]
    def __init__(self, signature, public_keys, message_hashes):
        self.signature = signature
        if len(public_keys) != len(message_hashes)
            raise ValidationError()
        self.public_keys = public_keys
        self.message_hashes = message_hashes
```
Then have
```
def verify_multiple(signed_message: SignedMessage): 
def verify_multiple_multiple(signed_messages: Sequence[SignedMessage]): 
```

tests/test_bls.py
ChihChengLiang: Removed randomness for now. Planning to rewrite bls tests with pytest fixtures and hypothesis.

py_ecc/bls/api.py
ChihChengLiang: Brought back groupby, since it's more like an optimization to save pairings for more general cases.

py_ecc/bls/api.py
asanso: @ChihChengLiang 

>I do desperately want to get rid of verify_multiple_multiple and rename it properly.

how about `batch_verify` ?

ChihChengLiang: Sounds good. We could also do "verify_multiple_aggregate_signatures" to match milagro's function name.

PR #70, opened 2019-05-23
Use SHA-256 instead of Keccak256

### LINKED ISSUE #68: Move to SHA256
* py-evm Version: 1.6.0
* OS: osx/linux/win
* Python Version (python --version):
* Environment (output of `pip freeze`):

### What is wrong?

ETH2.0 is using SHA256 instead of Keccak256 since v0.6.0 following this PR https://github.com/ethereum/eth2.0-specs/pull/779  
Py-ecc is still using Keccak256: https://github.com/ethereum/py_ecc/blob/master/py_ecc/bls/hash.py  
Py-ecc is used in BLS test generators etc.

### How can it be fixed

By changing hashing algo to SHA256 + new version release, so it could be used as dependency.

ralexstokes:
### What was wrong?

Fixes #68.

Updates the hash function so we can generate correct BLS tests in the eth2.0 specs/tests repo.

### How was it fixed?

Use the stdlib's SHA-256 instead of the Keccak256 import.

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,6 @@ setup(
     packages=find_packages(exclude=('tests', 'docs')),
     package_data={'py_ecc': ['py.typed']},
     install_requires=[
-        "eth-hash[pycryptodome]",
         "eth-typing>=2.1.0,<3.0.0",
         "eth-utils>=1.3.0,<2",
         "mypy-extensions>=0.4.1",
--- a/py_ecc/bls/hash.py
+++ b/py_ecc/bls/hash.py
@@ -1,12 +1,17 @@
-from eth_typing import Hash32
-from eth_hash.auto import keccak
+import hashlib
 from typing import Union
+
+from eth_typing import Hash32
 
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -45,7 +45,6 @@
     packages=find_packages(exclude=('tests', 'docs')),
     package_data={'py_ecc': ['py.typed']},
     install_requires=[
-        "eth-hash[pycryptodome]",
```
carver: Oh neat. Yay for dropping dependencies.

carver:
@ralexstokes Do you prefer squash merges or do you like to rebase/squash and then merge commit (I'm a fan of the latter)?

djrtwo:
Whatever you're a fan of @carver  :)

We're looking to merge this and get a minor version release out today. I can handle this unless you have a preferred workflow

carver:
> We're looking to merge this and get a minor version release out today. I can handle this unless you have a preferred workflow

Nope, no specific thing I need to be involved for. :+1:  Please update the changelog.

Issue #71, opened 2019-05-25
Wrapper for BLS python bindings in Chia Network


mhchia:
### What is wrong?
To address https://github.com/ethereum/py_ecc/issues/66, we can make use of faster implementations. Chia-Network has BLS signature implementation in C++, as well as the bindings for Python.

### How can it be fixed
To make use of the bindings, we need a wrapper for the essential functions in `bls.api`
- `sign`
- `privtopub`
- `verify`
- `aggregate_signatures`
- `aggregate_pubkeys`
- `aggregate_multiple`

mhchia:
cc @pipermerriam @carver @hwwhww @ChihChengLiang

ChihChengLiang:
Could we leave `bls/api.py` as is, and put the wrapper in somewhere like `bls_binding/api.py`?

For the license concern, I don't have experience in this either.

carver:
> [Chia-Network/bls-signatures](https://github.com/Chia-Network/bls-signatures) is under the license Apache 2.0. From my understanding, since we only import the library, from the rules in Apache 2.0, it seems not to affect our license. Is it correct? I don't have enough experience in the licenses. Please correct me if I'm wrong.

Yeah, no problem to add a dependency on an Apache-licensed library. It's pretty much only a problem with something like GPL-like licenses.

mhchia:
@carver That sounds great! Thank you for answering:)

PR #72, opened 2019-05-27
[WIP] Add wrapper for BLS from Chia Network

mhchia:
### What was wrong?
Fixes https://github.com/ethereum/py_ecc/issues/71. I will use "BLS chia" as the alias of Chia network's BLS implementation, along with its python bindings in the following paragraphs.

### How was it fixed?
- Implement functions `sign`, `privtopub`, `verify`, `aggregate_signatures`, `aggregate_pubkeys`, `aggregate_multiple` with BLS chia.
- Copy and modify the related tests from `tests/test_bls.py` to `tests/test_bls_chia.py` and test with those functions.
- Add a benchmark script `scripts/benchmark_api.py`, which is mostly inspired by https://github.com/ethereum/py_ecc/pull/69.

#### Problems
- The private key range in BLS chia seems to be different from py_ecc. Its range is [1, 2**32 - 1]. Therefore I currently workaround it through adding all private keys with `1`.
- Sometimes signature aggregation throws `RuntimeError: Caught an unknown exception!` if the source signatures are initialized with `blspy.Signature.from_bytes`. This issue is really hard to debug since no further information is printed from the C++ side. Currently, I workaround it through adding a `_sig_map: Dict[bytes, bls_chia.Signature] = {}`, which memoizes all signatures generated through `sign` and `aggregate_signatures`, to avoid the issue resulting from `Signature.from_bytes`. Ref: https://github.com/Chia-Network/bls-signatures/issues/54

#### Example benchmark
Run `number=100` times with `timeit`, with `20` keys and signatures aggregated. Need to further try the other parameters, and understand the meanings of the result.

```bash
# BLS chia
$ python py_ecc/scripts/benchmark_api.py bls_chia
Module py_ecc.bls_chia.api
privtopub: 0.012366111972369254
verify: 0.39142126898514107
aggregate_signatures: 2.8922514080186374
aggregate_pubkeys: 0.5968314119963907
verify_multiple: 4.292670964030549

# py_ecc itself
$ python py_ecc/scripts/benchmark_api.py bls
Module py_ecc.bls.api
privtopub: 1.719821373000741
verify: 121.40623104997212
aggregate_signatures: 31.77455569698941
aggregate_pubkeys: 1.140943904989399
verify_multiple: 763.4154342580005
```

To make use of the script with custom implementations:
- Implement `api.py` with functions `privtopub`, `verify`, `aggregate_signatures`, `aggregate_pubkeys`, `verify_multiple`.
- Put `api.py` under a package with your custom name, says "bls_another", under `py_ecc`, so `py_ecc/bls_another/api.py`.
- Run `python py_ecc/scripts/benchmark_api.py bls_another`.

#### Cute Animal Picture

### DIFF
--- a/py_ecc/bls_chia/__init__.py
+++ b/py_ecc/bls_chia/__init__.py
@@ -1,0 +1,8 @@
+from .api import (  # noqa: F401
+    aggregate_pubkeys,
+    aggregate_signatures,
+    privtopub,
+    sign,
+    verify,
+    verify_multiple,
+)
--- a/py_ecc/bls_chia/__init__.py
+++ b/py_ecc/bls_chia/__init__.py
@@ -1,8 +1,0 @@
-from .api import (  # noqa: F401
-    aggregate_pubkeys,
-    aggregate_signatures,
-    privtopub,
-    sign,
-    verify,
-    verify_multiple,
-)
--- a/py_ecc/bls_chia/api.py
+++ b/py_ecc/bls_chia/api.py
@@ -69,7 +69,7 @@ def aggregate_signatures(signatures: Sequence[BLSSignature]) -> BLSSignature:
     #   aggregating signatures
     global _sig_map
     signatures_chia = [
-        signature if signature in _sig_map else bls_chia.Signature.from_bytes(signature)
+        _sig_map[signature] if signature in _sig_map else bls_chia.Signature.from_bytes(signature)
         for signature in signatures
     ]
     aggregated_signature = bls_chia.Signature.aggregate(signatures_chia)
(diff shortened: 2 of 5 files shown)


mhchia:
CI fails due to the lack of CMake, but installing CMake and building result in much more time.(Though it can be mitigated with the cache in circle CI). I'm considering whether `test_bls_chia.py` should be added to the CI.

ralexstokes:
Worth thinking about experimenting w/ the code linked in #73

mhchia:
Agree! It looks pretty interesting. I think both C and pure Python worth our trying, too.

mhchia:
Moved to trinity in https://github.com/ethereum/trinity/pull/708

PR #75, opened 2019-07-05
Add tests for behaviors

### LINKED ISSUE #65: empty sums in BLS
### What is wrong?

https://github.com/ethereum/eth2.0-specs/pull/782

### How can it be fixed

This library should already satisfy the spec, no code change required. But can add tests to the empty aggregation cases.

### LINKED ISSUE #74: For arbitrary message, is empty signature valid for empty public key?
### What is wrong?

To py_ecc, if you sign a message with an empty public key and an empty signature, it verifies True. That seems wrong.

```python
In [10]: from py_ecc.bls import verify

In [11]: verify(b'\x56'*32, b'\xc0'+ b'\x00'*47, b'\xc0'+ b'\x00'*95, 0)
Out[11]: True
```

Arguments to return True
- It is a valid behavior for the pairing function
- It is just a special case of rogue key attack, we don't let validator input public key anyway.

Arguments to return False
- It's a scary gotcha to let an arbitrary message to be verified True, if not handled carefully.

ChihChengLiang:
### What was wrong?

Fix #74, add tests to confirm the behavior of the verifying empty public key and signature.

Fix #65 by the way, add tests to confirm the behavior of aggregating empty lists of public key and signature.

#### Cute Animal Picture

### DIFF
--- a/py_ecc/bls/constants.py
+++ b/py_ecc/bls/constants.py
@@ -4,6 +4,10 @@ from py_ecc.fields import (
 from py_ecc.optimized_bls12_381 import (
     field_modulus as q,
 )
+from eth_typing import (
+    BLSPubkey,
+    BLSSignature,
+)
 
 G2_cofactor = 305502333931268344200999753193121504214466019254188142667664032982267604182971884026507427359259977847832272839041616661285803823378372096355777062779109  # noqa: E501
 FQ2_order = q ** 2 - 1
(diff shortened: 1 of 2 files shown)


carver:
I'm on board with having `verify()` return `False` for the empty public key, if you'd like to get that folded in now. :+1:

hwwhww:
closing it now since empty aggregate is invalid in the new BLS standard.

PR #77, opened 2019-07-11
Introduce `Domain` type

### LINKED ISSUE #76: Introduce the `Domain` type
### What is wrong?

A new type was introduced for the BLS domain here: https://github.com/ethereum/eth2.0-specs/pull/1246

Right now, `py_ecc` takes the domain as an `int`.

### How can it be fixed

Add the `Domain` type and update the `bls` API.

ralexstokes:
### What was wrong?

Fixes #76.

### How was it fixed?

Adding a newtype of `Domain` from `bytes`.

#### Cute Animal Picture

### DIFF
--- a/tests/test_bls.py
+++ b/tests/test_bls.py
@@ -55,8 +55,8 @@ from py_ecc.optimized_bls12_381 import (
     ]
 )
 def test_get_x_coordinate(message_hash, domain):
-    x_coordinate = _get_x_coordinate(message_hash, domain)
     domain_in_bytes = domain.to_bytes(8, 'big')
+    x_coordinate = _get_x_coordinate(message_hash, domain_in_bytes)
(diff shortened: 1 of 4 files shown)


tests/test_bls.py
```diff
@@ -69,7 +69,8 @@ def test_hash_to_G2():
     message_hash = b'\x12' * 32
 
     domain_1 = 1
-    result_1 = hash_to_G2(message_hash, domain_1)
+    domain_in_bytes = domain_1.to_bytes(8, 'big')
```
ChihChengLiang: I think we can safely use little-endian here, so that we have the domain in little-endian everywhere.

ralexstokes: the spec still uses big-endian for this part -- i think the idea is to wait for the standardization before we change bls much more

Issue #73, opened 2019-05-28
Constant time hashing to the curve


ralexstokes:
### What is wrong?

For `bls12-381` we want to map messages to points on the curve. The current implementation uses an expensive "hash and check" method via `hash_to_g2`. It is desirable to have a more performant method of mapping a given message to the curve.

### How can it be fixed

Luckily, there was recent work on a constant time hashing method w/ implementation here:

https://github.com/kwantam/bls12-381_hash

There are some BLS standardization efforts under way that will likely involve making this method the canonical "hash to the curve". If this method makes its way into the standard, then we will definitely want to include it.

The linked repo has a C implementation and a pure Python implementation. We can add both, in the appropriate places, to get the functionality we want.

PR #79, opened 2019-09-16
Kirk baird patch 01

kirk-baird:
### What was wrong?

Updating the BLS `hash_to_G2` implementation to match the [standard](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve) and https://github.com/ethereum/eth2.0-specs/pull/1398

### How was it fixed?

Implement `hash_to_curve` for G2 which uses Boneh & Wahby method.

There are three parts to this:
- `hash_to_field` - convert a message to a point in the finite field.
- `map_to_curve` - SWU Map to 3-Isogenous Curve + 3-Isogeny Map to BLs12-318 G2
- `clear_cofactor` - ensure that the point is in the correct subfield.

Note: This is a Work In Progress PR to show progress and may not always be passing tests.

#### Cute Animal Picture


py_ecc/bls/hash.py
```diff
@@ -15,3 +18,39 @@ def hash_eth2(data: Union[bytes, bytearray]) -> Hash32:
...
+def hkdf_extract(salt: Union[bytes, bytearray], ikm: Union[bytes, bytearray]) -> bytes:
+    """
+    HKDF-Expand
+
+    https://tools.ietf.org/html/rfc5869
+    """
+    return hmac.new(salt, ikm, hashlib.sha256).digest()
+
+
+def hkdf_expand(prk: Union[bytes, bytearray], info: Union[bytes, bytearray], length: int) -> bytes:
```
pipermerriam: You could use [`typing.ByteString`](https://docs.python.org/3/library/typing.html#typing.ByteString) in place of `Union[bytes, bytearray]` (though that also includes `memoryview` types which I think should be fine.

py_ecc/bls/hash.py
```diff
@@ -15,3 +18,39 @@ def hash_eth2(data: Union[bytes, bytearray]) -> Hash32:
...
+    """
+    HKDF-Expand
+
+    https://tools.ietf.org/html/rfc5869
+    """
+    # n = cieling(length / HASH_LENGTH_BYTES)
+    n = length // HASH_LENGTH_BYTES
+    if n * HASH_LENGTH_BYTES < length:
```
pipermerriam: I think you can simplify this to:

```python
n = length // HASH_LENGTH_BYTES + length % 2
```

I think it might be a little more intuitive to read.

py_ecc/bls/hash.py
```diff
@@ -15,3 +18,39 @@ def hash_eth2(data: Union[bytes, bytearray]) -> Hash32:
...
+    okm = bytes(0)
+    previous = bytes(0)
+
+    for i in range(0, n):
+        # Concatenate (T(i) || info || i)
+        text = previous + info + bytes([i + 1])
+
+        # T(i + 1) = HMAC(T(i) || info || i)
+        previous = hmac.new(prk, text, hashlib.sha256).digest()
+        okm = okm + previous
```
pipermerriam: if you used a `bytearray` for the `okm` variable I think we could prevents some extra object copies by using `okm.extend(previous)`  Same for `previous`

py_ecc/bls/hash.py
```diff
@@ -15,3 +18,39 @@ def hash_eth2(data: Union[bytes, bytearray]) -> Hash32:
...
+
+    # Return first `length` bytes.
+    return okm[:length]
```
pipermerriam: Since we discard the end of this can we exit early from the loop above once the condition `okm >= length` is satisfied?

py_ecc/bls/utils.py
```diff
@@ -66,31 +73,62 @@ def modular_squareroot_in_FQ2(value: FQ2) -> FQ2:
...
-    )
+    Converts a message to a point in the finite field as defined here:
+    https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-04#section-5
+    """
+    m_prime = hkdf_extract(DST, message_hash)
+    e = []
+
+    for i in range(1, 3):
+        # Concatenate ("H2C" || I2OSP(ctr, 1) || I2OSP(i, 1))
+        info = b'H2C' + bytes([ctr])[:1] + bytes([i])[:1]
```
pipermerriam: I think the slices here are un-necessary.  I believe the following statement is equivalent:

```
info = b'H2C' + bytes([ctr]) + bytes([i])
```

py_ecc/fields/optimized_field_elements.py
```diff
@@ -182,6 +182,15 @@ def __repr__(self: T_FQ) -> str:
     def __int__(self: T_FQ) -> int:
         return self.n
 
+    def sgn0(self: Type[T_FQ]) -> int:
+        if self.n == 0:
+            return 0
+            neg = -self
```
pipermerriam: looks like your indentation got a little off here.

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,50 @@
...
+ISO_3_Z = FQ2([1, 1])
+P_MINUS_9_DIV_16 = 1001205140483106588246484290269935788605945006208159541241399033561623546780709821462541004956387089373434649096260670658193992783731681621012512651314777238193313314641988297376025498093520728838658813979860931248214124593092835  # noqa: E501
+SQRT_I = 1028732146235106349975324479215795277384839936929757896155643118032610843298655225875571310552543014690878354869257  # noqa: E501
+EV1 = 426061185569912361983521454249761337083267257081408520893788542915383290290183480196466860748572708974347122096641  # noqa: E501
+EV2 = 1288825690270127928862280779549770369920038885627059587892741294824265852728860506840371064237610802480748737721626  # noqa: E501
+
+positive_eigth_roots_of_unity = (
+    FQ2([1, 0]), FQ2([0, 1]), FQ2([SQRT_I, SQRT_I]), FQ2([SQRT_I, -SQRT_I]),
+)
+eighth_roots_of_unity = (
```
pipermerriam: conforming to the all-caps casing would be nice to maintain consistency.

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,138 @@
...
+            valid_root = True
+            result = temp1
+
+    return (valid_root, result)
+
+
+# Return the 4 positive 8th roots of unity in FQ2
+def positive_eigth_roots_of_unity_FQ2() -> Sequence[FQ2]:
+    roots = []
+    roots.append(FQ2([1, 0]))
```
pipermerriam: Thoughts on doing something like:

```python
return [
    FQ2([1, 0]),
    FQ2([0, 1]),
    ...
]
```

And since these seem to be constant values, maybe pre-computing at module load time and just returning the pre-computed/cached value?

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,138 @@
...
+    roots.append(FQ2([0, 1]))
+    roots.append(FQ2([SQRT_I, SQRT_I]))
+    roots.append(FQ2([SQRT_I, -SQRT_I]))
+    return roots
+
+
+# Setup the four positive roots of eta = sqrt(Z^3 * (-1)^(1 / 4))
+def positive_eta_roots() -> Sequence[FQ2]:
```
pipermerriam: For type hints we try to be *strict* about return types and forgiving about input types.  That would suggest that the return value here should be more specific like `Tuple[FQ2, ...]` (as we highly prefer tuples over lists as a way to avoid mutability ).

`Sequence` is typically reserved for argument types so that it's forgiving about allowing a broader range of inputs (like lists *and* tuples)

py_ecc/bls/hash.py
kirk-baird: Yep, I like that, thanks :)

kirk-baird: Actually on second though that wouldn't work. e.g.
`length = 34`, `HASH_LENGTH_BYTES = 32`
=> `n = 34 // 32 + 34 % 2 = 1`
But `n` should be 2.

pipermerriam: :man_facepalming:

py_ecc/bls/hash.py
kirk-baird: Yep, though `okm >= length` after `n` iterators of the loop.

The loop could be changed to the condition `while len(okm) < length`
Then we would not have to calculate `n`. However since it is following [RFC5869](https://tools.ietf.org/html/rfc5869) and that is the way it's mentioned there I think it'd be better to leave it with `n`.

py_ecc/bls/utils.py
kirk-baird: I added these as insurance that `bytes([ctr])[:1] + bytes([i])[:1]` can oly ever be 1 byte.

You're right about `[i][:1]` that can be removed since i is either 1 or 2. However for `ctr` its type is `int`so to prevent accidental (or deliberate) misuse I thought it appropriate to ensure it is a slice of length 1. If you think this is overkill I'm happy to remove the slice.

Unless there is a way in the function to change `ctr` type to a single byte or u8?
`def hash_to_base_FQ2(message_hash: Hash32, ctr: int) -> FQ2:`

py_ecc/fields/optimized_field_elements.py
kirk-baird: Thanks, it sure is!

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: Yep that makes sense to me :)

py_ecc/optimized_bls12_381/constants.py
kirk-baird: Done :)

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: Yep that's an old function of mine that's now computer at module load time, so I'll remove that and update the call to it.

py_ecc/bls/utils.py
pipermerriam:
```python
----> 1 bytes([256])

ValueError: bytes must be in range(0, 256)
```

I don't think it's possible for `bytes([ctr])` to be longer than a single byte.

py_ecc/bls/hash.py
kirk-baird: Sounds good, I'm happy to switch it to `ByteString` and include `memoryview`

py_ecc/bls/hash.py
kirk-baird: Yep I agree :)

py_ecc/bls/utils.py
kirk-baird: You're right I'll remove the slicing :)

kirk-baird:
Update: Sorry I've been off this for awhile I've been working on a couple of other things + DevCon but I'll  finish implementing it from now.

py_ecc/bls/hash.py
kirk-baird: I remember why I choose `Union[bytes, bytearray]` now and that is because the underlying `sha256` function takes `Union[bytes, bytearray]` so I will switch it back to `Union`

kirk-baird:
Although this branch should now be usable there are a few things left todo:
- Increase test coverage of `hash_to_curve`
- Optimise `clear_cofactor` if it is allowed as per patent
- This should be constant time therefore solving #73 but worth bench marking to check
- Update links once bls standard drafts are finalised
- `PopVerify` and `PopProve` functions

Also worth noting that compression/decompression will need to be updated (currently looks like they will be made to match zcash).

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,49 @@
+from py_ecc.fields import (
+    optimized_bls12_381_FQ2 as FQ2,
+)
+
+# Ciphersuite BLS12381G2-SHA256-SSWU-RO paramters
+ISO_3_A = FQ2([0, 240])
+ISO_3_B = FQ2([1012, 1012])
+ISO_3_Z = FQ2([1, 1])
+P_MINUS_9_DIV_16 = 1001205140483106588246484290269935788605945006208159541241399033561623546780709821462541004956387089373434649096260670658193992783731681621012512651314777238193313314641988297376025498093520728838658813979860931248214124593092835  # noqa: E501
+SQRT_I = 1028732146235106349975324479215795277384839936929757896155643118032610843298655225875571310552543014690878354869257  # noqa: E501
```
pipermerriam: I have some discomfort from having all of these constants floating around.  Is there a formal/canonical spec that we can include a link to so that it's easy to reference/cross-check these constants are indeed the values they are supposed to be?

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,49 @@
...
+EV1 = 426061185569912361983521454249761337083267257081408520893788542915383290290183480196466860748572708974347122096641  # noqa: E501
+EV2 = 1288825690270127928862280779549770369920038885627059587892741294824265852728860506840371064237610802480748737721626  # noqa: E501
+
+POSITIVE_EIGTH_ROOTS_OF_UNITY = (
+    FQ2([1, 0]), FQ2([0, 1]), FQ2([SQRT_I, SQRT_I]), FQ2([SQRT_I, -SQRT_I]),
+)
+
+# X Numerator
+ISO_3_K_1_0 = FQ2([889424345604814976315064405719089812568196182208668418962679585805340366775741747653930584250892369786198727235542, 889424345604814976315064405719089812568196182208668418962679585805340366775741747653930584250892369786198727235542])  # noqa: E501
```
pipermerriam: The number inlined here seems to be repeated twice.  Can we set it to a constant to avoid repeating it here?

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,49 @@
...
+
+# X Denominator
+ISO_3_K_2_0 = FQ2([0, 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559715])  # noqa: E501
+ISO_3_K_2_1 = FQ2([12, 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559775])  # noqa: E501
+ISO_3_K_2_2 = FQ2.one()
+ISO_3_K_2_3 = FQ2.zero()
+ISO_3_X_DENOMINATOR = (ISO_3_K_2_0, ISO_3_K_2_1, ISO_3_K_2_2, ISO_3_K_2_3)
+
+# Y Numerator
+ISO_3_K_3_0 = FQ2([3261222600550988246488569487636662646083386001431784202863158481286248011511053074731078808919938689216061999863558, 3261222600550988246488569487636662646083386001431784202863158481286248011511053074731078808919938689216061999863558])  # noqa: E501
```
pipermerriam: Same here with this literal value being repeated.

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,49 @@
...
+ISO_3_K_3_1 = FQ2([0, 889424345604814976315064405719089812568196182208668418962679585805340366775741747653930584250892369786198727235518])  # noqa: E501
+ISO_3_K_3_2 = FQ2([2668273036814444928945193217157269437704588546626005256888038757416021100327225242961791752752677109358596181706524, 1334136518407222464472596608578634718852294273313002628444019378708010550163612621480895876376338554679298090853263])  # noqa: E501
+ISO_3_K_3_3 = FQ2([2816510427748580758331037284777117739799287910327449993381818688383577828123182200904113516794492504322962636245776, 0])  # noqa: E501
+ISO_3_Y_NUMERATOR = (ISO_3_K_3_0, ISO_3_K_3_1, ISO_3_K_3_2, ISO_3_K_3_3)
+
+# Y Denominator
+ISO_3_K_4_0 = FQ2([4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559355, 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559355])  # noqa: E501
```
pipermerriam: and here

py_ecc/fields/optimized_field_elements.py
```diff
@@ -363,6 +372,24 @@ def __ne__(self: T_FQP, other: T_FQP) -> bool:     # type: ignore # https://gith
...
+        for x_i in reversed(self.coeffs):
+            sign_i = 0
+            if isinstance(x_i, int):
+                if x_i == 0:
+                    sign_i = 0
+                elif (-x_i % self.field_modulus) > (x_i % self.field_modulus):
+                    sign_i = 1
+                else:
+                    sign_i = -1
+            else:
```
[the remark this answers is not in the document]
kirk-baird: Yep sounds like good practice, I'll update it.

py_ecc/optimized_bls12_381/constants.py
kirk-baird: There are soo many constants! Unfortunately they are necessary for the isogeny mapping.

Most of them come from
BLS General Constants: https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-04#section-8.7
ISO constants: https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-04#appendix-C.2

py_ecc/optimized_bls12_381/constants.py
kirk-baird: Yep can do.

kirk-baird:
Update:

[Version 5](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/releases/tag/draft-irtf-cfrg-hash-to-curve-05) of the hash-to-curve spec is out.

I will update the code soon to match the new version. This is expected to be the last breaking change so hopefully won't be too long before we'll be able to merge.

py_ecc/bls/api.py
```diff
@@ -52,17 +49,18 @@ def privtopub(k: int) -> BLSPubkey:
 
 def verify(message_hash: Hash32,
            pubkey: BLSPubkey,
-           signature: BLSSignature,
-           domain: Domain) -> bool:
+           signature: BLSSignature) -> bool:
+    signature_point = signature_to_G2(signature)
+    if not is_inf(multiply(signature_point, G2_COFACTOR)):
```
hwwhww: It should be `curve_order` from https://github.com/ethereum/py_ecc/blob/c74aca90b7905a1ec9dd1454dc992d5c6f4f28b0/py_ecc/optimized_bls12_381/optimized_curve.py#L20 instead of `G2_COFACTOR`.
```suggestion
    if not is_inf(multiply(signature_point, curve_order)):
```

py_ecc/bls/api.py
```diff
@@ -99,6 +96,10 @@ def verify_multiple(pubkeys: Sequence[BLSPubkey],
             )
         )
 
+    signature_point = signature_to_G2(signature)
+    if not is_inf(multiply(signature_point, G2_COFACTOR)):
```
hwwhww: ditto.

py_ecc/bls/api.py
```diff
@@ -34,14 +32,13 @@
     pubkey_to_G1,
     signature_to_G2,
 )
-
+from .constants import G2_COFACTOR
```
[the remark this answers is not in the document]
kirk-baird: Sure does!

py_ecc/bls/utils.py
```diff
@@ -55,41 +59,74 @@ def modular_squareroot_in_FQ2(value: FQ2) -> FQ2:
...
-    return x_coordinate
 
+def hash_to_curve_G2(message_hash: Hash32) -> G2Uncompressed:
+    u0 = hash_to_base_FQ2(message_hash, 0)
+    u1 = hash_to_base_FQ2(message_hash, 1)
+    q0 = map_to_curve_G2(u0)
+    q1 = map_to_curve_G2(u1)
+    r = add(q0, q1)
+    p = clear_cofactor_G2(r)
+    return p
```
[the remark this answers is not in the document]
kirk-baird: Yep, I'll combine these :)

py_ecc/bls/api.py
kirk-baird: Ooh Yes good pickup!!

kirk-baird:
I happy for this to be merged now :)

Some final notes:
- I have based this from against the [hash-to-curve standard version 5](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-05). 
- I've made some of the functions constant time such as `sqrt_division_FQ2` and `optimized_swu_G2`. We could make some pretty decent speed ups if these are not made to be constant time (constant time hashing is not necessary for Ethereum).
- The tests are based off input to the [reference implementation](https://github.com/kwantam/bls_sigs_ref/tree/master/python-impl)

py_ecc/bls/utils.py
```diff
@@ -55,41 +59,74 @@ def modular_squareroot_in_FQ2(value: FQ2) -> FQ2:
...
+    Hash To Base for FQ2
 
-    # Test candidate y coordinates until a one is found
-    while 1:
-        y_coordinate_squared = x_coordinate ** 3 + FQ2([4, 4])  # The curve is y^2 = x^3 + 4(i + 1)
-        y_coordinate = modular_squareroot_in_FQ2(y_coordinate_squared)
-        if y_coordinate is not None:  # Check if quadratic residue found
-            break
-        x_coordinate += FQ2([1, 0])  # Add 1 and try again
+    Converts a message to a point in the finite field as defined here:
```
hwwhww: Nitpick:
```suggestion
    Convert a message to a point in the finite field as defined here:
```

py_ecc/bls/hash.py
```diff
@@ -1,8 +1,11 @@
 import hashlib
+import hmac
 from typing import Union
 
 from eth_typing import Hash32
 
+from .constants import HASH_LENGTH_BYTES
+
 
 def hash_eth2(data: Union[bytes, bytearray]) -> Hash32:
```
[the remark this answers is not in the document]
kirk-baird: Nope, I'll remove it.

py_ecc/bls/utils.py
```diff
@@ -55,41 +59,74 @@ def modular_squareroot_in_FQ2(value: FQ2) -> FQ2:
...
+    """
+    m_prime = hkdf_extract(DST, message_hash + b'\x00')
+    info_pfx = b'H2C' + bytes([ctr])
+    e = []
 
-    return multiply(
-        (x_coordinate, y_coordinate, FQ2([1, 0])),
-        G2_cofactor,
-    )
+    for i in range(1, 3):
```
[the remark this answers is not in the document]
kirk-baird: Sure thing :)

py_ecc/bls/utils.py
```diff
@@ -55,41 +59,74 @@ def modular_squareroot_in_FQ2(value: FQ2) -> FQ2:
...
+
+
+def map_to_curve_G2(u: FQ2) -> G2Uncompressed:
+    """
+    Map To Curve for G2
+
+    First, convert FQ2 point to a point on the 3-Isogeny curve.
+    SWU Map: https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-05#section-6.6.2
+
+    Second, map 3-Isogeny curve to BLS381-G2 curve.
```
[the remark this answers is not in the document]
kirk-baird: Haha yep, I'll fix that for thoroughness. As an FYI BLS12-381 is sometimes abbreviated to BLS381.

py_ecc/fields/optimized_field_elements.py
```diff
@@ -182,6 +182,14 @@ def __repr__(self: T_FQ) -> str:
     def __int__(self: T_FQ) -> int:
         return self.n
 
+    def sgn0(self: T_FQ) -> int:
```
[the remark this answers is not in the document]
kirk-baird: Yep can do :)

py_ecc/fields/optimized_field_elements.py
```diff
@@ -182,6 +182,14 @@ def __repr__(self: T_FQ) -> str:
...
+        if self.n == 0:
+            return 0
```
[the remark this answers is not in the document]
kirk-baird: Oh yep that was change between version 4 and 5! 
Good pickup.

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,58 @@
...
+ISO_3_A = FQ2([0, 240])
+ISO_3_B = FQ2([1012, 1012])
+ISO_3_Z = FQ2([-2, -1])
+P_MINUS_9_DIV_16 = 1001205140483106588246484290269935788605945006208159541241399033561623546780709821462541004956387089373434649096260670658193992783731681621012512651314777238193313314641988297376025498093520728838658813979860931248214124593092835  # noqa: E501
+
+EV1 = 1015919005498129635886032702454337503112659152043614931979881174103627376789972962005013361970813319613593700736144  # noqa: E501
+EV2 = 1244231661155348484223428017511856347821538750986231559855759541903146219579071812422210818684355842447591283616181  # noqa: E501
+EV3 = 1646015993121829755895883253076789309308090876275172350194834453434199515639474951814226234213676147507404483718679  # noqa: E501
+EV4 = 1637752706019426886789797193293828301565549384974986623510918743054325021588194075665960171838131772227885159387073  # noqa: E501
+ETAS = [FQ2([EV1, EV2]), FQ2([-EV2, EV1]), FQ2([EV3, EV4]), FQ2([-EV4, EV3])]
```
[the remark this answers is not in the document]
kirk-baird: Eta from the [Greek symbol](https://en.wikipedia.org/wiki/Eta) as that's how it appears in the papers. Etas cause there are a 4.

EV was short for Eta Vector, they could be named anything really.

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+def sqrt_division_FQ2(u: FQ2, v: FQ2) -> Tuple[bool, FQ2]:
+    temp1 = u * v ** 7
+    temp2 = temp1 * v ** 8
+
+    # sqrt_candidate =  uv^7 * (uv^15)^((p - 9) / 16)
+    sqrt_candidate = temp2 ** P_MINUS_9_DIV_16
+    sqrt_candidate = sqrt_candidate * temp1
+
+    # Verify sqrt_candidate is a valid root
+    valid_root = False
```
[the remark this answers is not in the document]
kirk-baird: Yep that'll be more readable.

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+    result = sqrt_candidate
+    roots = POSITIVE_EIGTH_ROOTS_OF_UNITY
+    for root in roots:
+        # Valid if (root * sqrt_candidate)^2 * v - u == 0
+        temp1 = (root * sqrt_candidate)
+        temp2 = temp1 ** 2 * v - u
+        if temp2 == FQ2.zero() and not valid_root:
+            valid_root = True
+            result = temp1
```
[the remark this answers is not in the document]
kirk-baird: This is where I deliberate left it as a constant time implementation.
We could get some speed ups here if we were to remove those checks. 
Ethereum does not need constant time hash to curve cause the messages are public so I'm happy to add the break statements in there if you would like?

py_ecc/bls/utils.py
kirk-baird: Sure :)

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+
+# Optimized SWU Map - FQ2 to G2': y^2 = x^3 + 240i * x + 1012 + 1012i
+# Found in Section 4 of https://eprint.iacr.org/2019/403
+def optimized_swu_G2(t: FQ2) -> Tuple[FQ2, FQ2, FQ2]:
+    t2 = t ** 2
+    temp = ISO_3_Z * t2
+    temp = temp + temp ** 2
+    denominator = -(ISO_3_A * temp)  # -a(Z * t^2 + Z^2 * t^4)
+    temp = temp + FQ2.one()
+    numerator = ISO_3_B * temp  # b(Z * t^2 + Z^2 * t^4 + + 1)
```
ChihChengLiang:
```suggestion
    numerator = ISO_3_B * temp  # b(Z * t^2 + Z^2 * t^4 + 1)
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+    if t.sgn0_be() != y.sgn0_be():
+        y = -y
+
+    y = y * denominator
+
+    return (numerator, y, denominator)
+
+
+# Square Root Division
+# Return: uv^7 * (uv^15)^((p - 9) / 16) * root of unity
```
ChihChengLiang: Is it p^2 here?
```suggestion
# Return: uv^7 * (uv^15)^((p^2 - 9) / 16) * root of unity
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+# If valid square root is found return true, else false
+def sqrt_division_FQ2(u: FQ2, v: FQ2) -> Tuple[bool, FQ2]:
+    temp1 = u * v ** 7
+    temp2 = temp1 * v ** 8
+
+    # sqrt_candidate =  uv^7 * (uv^15)^((p - 9) / 16)
```
ChihChengLiang: Is it p^2 here?
```suggestion
    # sqrt_candidate =  uv^7 * (uv^15)^((p^2 - 9) / 16)
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+    sqrt_candidate = temp2 ** P_MINUS_9_DIV_16
+    sqrt_candidate = sqrt_candidate * temp1
+
+    # Verify sqrt_candidate is a valid root
+    is_valid_root = False
+    result = sqrt_candidate
+    roots = POSITIVE_EIGTH_ROOTS_OF_UNITY
+    for root in roots:
+        # Valid if (root * sqrt_candidate)^2 * v - u == 0
+        temp1 = (root * sqrt_candidate)
```
ChihChengLiang: It looks like `temp1` is the real `sqrt_candidate`. Should we rename `sqrt_candidate` to `gamma` and
 `temp1` to `sqrt_candidate`?

py_ecc/optimized_bls12_381/optimized_swu.py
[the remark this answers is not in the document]
pipermerriam: Not sure if this instinct is right but maybe lets leave this as constant time since that seems like a *safe* default.

We can track this in an issue as an option for performance improvements if needed.

py_ecc/bls/hash.py
```diff
@@ -1,17 +1,41 @@
...
+    """
+    HKDF-Expand
 
-    Note: it's a placeholder and we aim to migrate to a S[T/N]ARK-friendly hash function in
-    a future Ethereum 2.0 deployment phase.
+    https://tools.ietf.org/html/rfc5869
     """
-    return Hash32(hashlib.sha256(data).digest())
+    # n = cieling(length / HASH_LENGTH_BYTES)
+    n = length // HASH_LENGTH_BYTES
```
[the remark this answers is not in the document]
kirk-baird: Yep that's a better way :)

py_ecc/optimized_bls12_381/constants.py
```diff
@@ -0,0 +1,58 @@
...
+ISO_3_K_4_0_VAL = 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559355  # noqa: E501
+ISO_3_K_4_0 = FQ2([ISO_3_K_4_0_VAL, ISO_3_K_4_0_VAL])  # noqa: E501
+ISO_3_K_4_1 = FQ2([0, 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559571])  # noqa: E501
+ISO_3_K_4_2 = FQ2([18, 4002409555221667393417789825735904156556882819939007885332058136124031650490837864442687629129015664037894272559769])  # noqa: E501
+ISO_3_K_4_3 = FQ2.one()
+ISO_3_Y_DENOMINATOR = (ISO_3_K_4_0, ISO_3_K_4_1, ISO_3_K_4_2, ISO_3_K_4_3)
+
+ISO_3_MAP_COEFFICIENTS = (ISO_3_X_NUMERATOR, ISO_3_X_DENOMINATOR, ISO_3_Y_NUMERATOR, ISO_3_Y_DENOMINATOR)  # noqa: E501
+
+# 3 * (z**2 -1) * g2_cofactor
```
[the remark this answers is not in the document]
kirk-baird: Yep that's `h_eff` from [8.9.2](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-05)

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: Oops, fixed :)

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: Yep it's `p^2` though since we are in modulus p: `p^2 = p`.
I'll update it to avoid confusion though :)

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: Yep I'll update these accordingly

py_ecc/optimized_bls12_381/optimized_swu.py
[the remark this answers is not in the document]
kirk-baird: Sure I'll leave it as is and write-up an issue for it now.

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+    gamma = gamma * temp1
+
+    # Verify there is a valid root
+    is_valid_root = False
+    result = gamma
+    roots = POSITIVE_EIGTH_ROOTS_OF_UNITY
+    for root in roots:
+        # Valid if (root * gamma)^2 * v - u == 0
+        sqrt_candidate = (root * gamma)
+        temp2 = temp1 ** 2 * v - u
```
ChihChengLiang: The test is not passing because of this line
```suggestion
        temp2 = sqrt_candidate ** 2 * v - u
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+
+    if not success:
+        numerator = numerator * t2 * ISO_3_Z
+
+    if t.sgn0_be() != y.sgn0_be():
+        y = -y
+
+    y = y * denominator
+
+    return (numerator, y, denominator)
```
ChihChengLiang: Question: The mapping from Jacobian coordinates to the affine coordinates should be `(X, Y, Z) -> (X/ Z**2, Y / Z**3 )`

According to section 4.4 in the paper, the return value should look like 

```
# Omitting xi t^2, t^3, sign(t) here
(numerator * denominator, sqrt_candidate * denominator**3, denominator)
```

tests/test_bls_hash_to_G2.py
```diff
@@ -0,0 +1,109 @@
...
+        FQ2([int('119a71e0d20489cf8c5f82c51e879e7b344e53307b53be650df7f3f04907b75b71fdafe26e8d4e14e603440b09efe6f3', 16), int('e4ea193377da29537e8fe6f6f631adff10afaef2ea1eb2107d30d97358c1a19975e0a8bb62650ff90447cf5b3719c1d', 16)]),
+        FQ2([int('1142f2e077cc4230ee3cf07565ee626141ea9b86a79a0422d7f0e84c281ca09c5bbbe95f21e51285618c81d6dfda943d', 16), int('15e4da056552343eb519b3087962521d112c7e307d731373e8f1c72415306ccbc3c14fc6d68d61d2feeda3ea2e7729f', 16)])),
+    ]
+)
+def test_hash_to_G2(msg, x, y):
+    point = hash_to_G2(msg)
+
+    # Affine
+    result_x = point[0] / point[2] # X / Z
+    result_y = point[1] / point[2] # Y / Z
```
ChihChengLiang: Should it be 
```python
X, Y, Z = hash_to_G2(msg)

result_x = X / Z**2
result_y = Y / Z**3
```

tests/test_bls_hash_to_G2.py
```diff
@@ -0,0 +1,109 @@
...
+
```
ChihChengLiang: Maybe we can add a `assert is_on_curve(point, b2)` check here.

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+def iso_map_G2(x: FQ2, y: FQ2, z: FQ2) -> Optimized_Point3D[FQ2]:
+    # x-numerator, x-denominator, y-numerator, y-denominator
+    mapped_values = [FQ2.zero(), FQ2.zero(), FQ2.zero(), FQ2.zero()]
+    z_powers = [z, z ** 2, z ** 3]
+
+    # Horner Polynomial Evaluation
+    for (i, k_i) in enumerate(ISO_3_MAP_COEFFICIENTS):
+        mapped_values[i] = k_i[-1:][0]
+        for (j, k_i_j) in enumerate(reversed(k_i[:-1])):
+            mapped_values[i] = mapped_values[i] * x + z_powers[j] * k_i_j
```
ChihChengLiang: Should raise the power of x.
```suggestion
            mapped_values[i] = mapped_values[i] * x ** j + z_powers[j] * k_i_j
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+
+    mapped_values[2] = mapped_values[2] * y  # y-numerator * y
+    mapped_values[3] = mapped_values[3] * z  # y-denominator * z
+
+    z_G2 = mapped_values[1] * mapped_values[3]  # x-denominator * y-denominator
+    x_G2 = mapped_values[0] * mapped_values[3]  # x-numerator * y-denominator
```
ChihChengLiang: Should have z_G2 here
```suggestion
    x_G2 = mapped_values[0] * mapped_values[3] * z_G2  # x-numerator * y-denominator
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+    y_G2 = mapped_values[1] * mapped_values[2]  # y-numerator * x-denominator
```
ChihChengLiang: Should have z_G2 here too
```suggestion
    y_G2 = mapped_values[1] * mapped_values[2] * z_G2**2 # y-numerator * x-denominator
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,120 @@
...
+    success_2 = False
+    etas = ETAS
+    for (i, eta) in enumerate(etas):
+        # Valid solution if (eta * sqrt_candidate(x1)) ** 2 * v - u == 0
+        temp1 = eta * sqrt_candidate
+        temp1 = temp1 ** 2 * v - u
+        if temp1 == FQ2.zero() and not success and not success_2:
+            y = sqrt_candidate * eta
+            success_2 = True
+        elif i == 3 and not success and not success_2:
```
ChihChengLiang: nitpick: Python has a `else` feature that allows us to check conditions after the `for` loop
```python
for eta in ETAS:
    ...
else:
    if not success and not success_2:
        ...
```

py_ecc/optimized_bls12_381/optimized_swu.py
ChihChengLiang: Ah ... sorry, please ignore this one

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: That's a good question! There are two ways to make projective (Jacobian) coordinates.
- (X / Z, Y / Z)
- (X / Z^2, Y / Z^3)

As the `py_ecc` library uses `(X / Z, Y / Z)` hence the final output must be in that form. Thus, I've had to diverge from that paper such that we have end up with our projective form.

If we were to switch to `(X / Z^2, Y / Z^3)` we would have to re-write most of the library. The `(X / Z^2, Y / Z^3)` form is a little slower but has some protection against certain side channel attacks.

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: This also relates to the projective form. Since we have `(X / Z, Y / Z)` we have designed X such that 
`X / Z = x-numerator / x-denominator`

This differs from the paper which has `(X / Z^2, Y / Z^3)` hence they need
`X / Z^2 = x-numerator / x-denominator`

Similarly for Y.

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: That's a nice feature :)

py_ecc/optimized_bls12_381/optimized_swu.py
kirk-baird: Ahh silly mistake, thanks :)

py_ecc/optimized_bls12_381/optimized_swu.py
ChihChengLiang: Yep, sticking to`(X / Z, Y / Z)` sounds better.

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,121 @@
...
+    ISO_3_MAP_COEFFICIENTS,
+    POSITIVE_EIGTH_ROOTS_OF_UNITY,
+)
+
+
+# Optimized SWU Map - FQ2 to G2': y^2 = x^3 + 240i * x + 1012 + 1012i
+# Found in Section 4 of https://eprint.iacr.org/2019/403
+def optimized_swu_G2(t: FQ2) -> Tuple[FQ2, FQ2, FQ2]:
+    t2 = t ** 2
+    temp = ISO_3_Z * t2
```
ChihChengLiang: This `iso_3_z_t2` can be reused and save some multiplication
```suggestion
    iso_3_z_t2 = ISO_3_Z * t2
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,121 @@
...
+    # Attempt y = sqrt(u / v)
+    (success, sqrt_candidate) = sqrt_division_FQ2(u, v)
+    y = sqrt_candidate
+
+    # Handle case where (u / v) is not square
+    # sqrt_candidate(x1) = sqrt_candidate(x0) * t^3
+    sqrt_candidate = sqrt_candidate * t ** 3
+
+    # u(x1) = Z^3 * t^6 * u(x0)
+    u = (ISO_3_Z * t2) ** 3 * u
```
ChihChengLiang: here's an `iso_3_z_t2` reuse (1)
```suggestion
    u = iso_3_z_t2 ** 3 * u
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,121 @@
...
+    success_2 = False
+    etas = ETAS
+    for (i, eta) in enumerate(etas):
```
ChihChengLiang: unused i
```suggestion
    for eta in etas:
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,121 @@
...
+        if temp1 == FQ2.zero() and not success and not success_2:
+            y = sqrt_candidate * eta
+            success_2 = True
+    else:
+        if not success and not success_2:
+            # Unreachable
+            raise Exception("Hash to Curve - Optimized SWU failure")
+
+    if not success:
+        numerator = numerator * t2 * ISO_3_Z
```
ChihChengLiang: `iso_3_z_t2` reuse (2)
```suggestion
        numerator = numerator * iso_3_z_t2
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,121 @@
...
+    # sqrt_candidate(x1) = sqrt_candidate(x0) * t^3
+    sqrt_candidate = sqrt_candidate * t ** 3
+
+    # u(x1) = Z^3 * t^6 * u(x0)
+    u = (ISO_3_Z * t2) ** 3 * u
+    success_2 = False
+    etas = ETAS
+    for (i, eta) in enumerate(etas):
+        # Valid solution if (eta * sqrt_candidate(x1)) ** 2 * v - u == 0
+        temp1 = eta * sqrt_candidate
```
ChihChengLiang: `eta_sqrt_candidate` can be reused
```suggestion
        eta_sqrt_candidate = eta * sqrt_candidate
```

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -0,0 +1,121 @@
...
+        temp1 = temp1 ** 2 * v - u
+        if temp1 == FQ2.zero() and not success and not success_2:
+            y = sqrt_candidate * eta
```
ChihChengLiang: Reuse `eta_sqrt_candidate` here
```suggestion
            y = eta_sqrt_candidate
```

kirk-baird:
> My final review and some nitpick on reusing variables.

Looks good, I'll add all of these!

PR #83, opened 2019-12-19
Implements All G2 ciphersuites from the BLS Standards

CarlBeek:
### What was wrong?

0. The BLS aspects of `py_ecc` did not implement the interfaces specified by the BLS standards.
1. Only one of the ciphersuites was implemented. Now all three of the G2 ciphersuites are present.

### How was it fixed?

By refactoring how the BLS API was implemented by building modular components that allowed for all the ciphersuites to be implemented.

#### Cute Animal Picture

I saw a few baby warthogs in the bush last weekend.
image


CarlBeek:
I'd appreciate some assistance with the interfaces. Specifically, the way the cipher suites are implemented. Currently each ciphersuite is a file, but this leads to unused imports as certain functions needn't be redefined by the specific ciphersuite. The most obvious solution is to make each ciphersuite a folder with an `__init__.py` which would neaten up things at the cost of a tedious folder structure. Thoughts?

hwwhww:
IMO it seems natural to abstract it.
1. Implement a base interface class with `abc`, so we will have `class BaseCiphersuites(ABC)` to define the interfaces.
2. Three subclass implementations in different files.

CarlBeek:
@hwwhww, I agree classes make this much cleaner. I was avoiding them to be more zen 🧘‍♂️but it backfired.

What I don't like about what I've done now is that `Sign` etc are not static methods because they need access to `self.DST`.  This means that it is necessary to instantiate the class before use which is ugly and a bad interface. Do you have a good idea on how to fix it?

py_ecc/bls/ciphersuites.py
```diff
@@ -0,0 +1,187 @@
...
+                message_point = hash_to_G2(message, DST)
+                accumulator *= pairing(message_point, pubkey_point, final_exponentiate=False)
+            accumulator *= pairing(signature_point, neg(G1), final_exponentiate=False)
+            return final_exponentiate(accumulator) == FQ12.one()
+
+        except (ValidationError, ValueError, AssertionError):
+            return False
+
+    def Sign(self, SK: int, message: bytes) -> BLSSignature:
+        return self._CoreSign(SK, message, self.DST)
```
hwwhww: > What I don't like about what I've done now is that `Sign` etc are not static methods because they need access to `self.DST`.  This means that it is necessary to instantiate the class before use which is ugly and a bad interface. Do you have a good idea on how to fix it?

I think you can:
1. Make `DST` a class variable.
2. Change the APIs which need `DST` to `classmethod`s.

```python
class BaseG2Ciphersuite(ABC):
    DST = b''

    @classmethod
    def Sign(cls, SK: int, message: bytes) -> BLSSignature:
        return self._CoreSign(SK, message, cls.DST)
```

py_ecc/bls/ciphersuites.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def Verify(self, PK: BLSPubkey, message: bytes, signature: BLSSignature) -> bool:
+        return self._CoreVerify(PK, message, signature, self.DST)
+
+    def AggregateVerify(self, pairs: Iterable[Tuple[BLSPubkey, bytes]],
+                        signature: BLSSignature) -> bool:
+        pass
```
hwwhww: This is where we can add a `@abc.abstractmethod`.
```suggestion
    @abstractmethod
    def AggregateVerify(self, pairs: Iterable[Tuple[BLSPubkey, bytes]],
                        signature: BLSSignature) -> bool:
        ...
```

So that if a subclass of `BaseG2Ciphersuite` doesn't implement `AggregateVerify`, it will scream.

py_ecc/bls/ciphersuites.py
CarlBeek: Thanks, works like a charm!

py_ecc/bls/ciphersuites.py
```diff
@@ -0,0 +1,193 @@
...
+            signature_point = signature_to_G2(signature)
+            accumulator = add(accumulator, signature_point)
+        return G2_to_signature(accumulator)
+
+    @staticmethod
+    def _CoreAggregateVerify(pairs: Iterable[Tuple[BLSPubkey, bytes]],
+                             signature: BLSSignature, DST: bytes) -> bool:
+        try:
+            signature_point = signature_to_G2(signature)
+            accumulator = FQ12([1] + [0] * 11)
```
hwwhww: we can use `FQ12.one()`.
```suggestion
            accumulator = FQ12.one()
```

py_ecc/bls/ciphersuites.py
```diff
@@ -0,0 +1,193 @@
+from typing import (
+    Iterable,
```
hwwhww: We can be replaced all `Iterable`s with `Sequence`s in this PR.

py_ecc/bls/point_compression.py
```diff
@@ -177,18 +77,26 @@ def decompress_G1(z: G1Compressed) -> G1Uncompressed:
...
-
-
-def pubkey_to_G1(pubkey: BLSPubkey) -> G1Uncompressed:
-    z = big_endian_to_int(pubkey)
-    return decompress_G1(G1Compressed(z))
-
 #
 # G2
 #
+def modular_squareroot_in_FQ2(value: FQ2) -> Optional[FQ2]:
```
hwwhww: Nice catch!

py_ecc/bls/ciphersuites.py
```diff
@@ -0,0 +1,193 @@
...
+        return cls._CoreSign(SK, message, cls.DST)
+
+    @classmethod
+    def Verify(cls, PK: BLSPubkey, message: bytes, signature: BLSSignature) -> bool:
+        return cls._CoreVerify(PK, message, signature, cls.DST)
+
+    @abc.abstractclassmethod
+    def AggregateVerify(cls, pairs: Iterable[Tuple[BLSPubkey, bytes]],
+                        signature: BLSSignature) -> bool:
+        pass
```
hwwhww: nitpick
```suggestion
        ...
```

py_ecc/bls/ciphersuites.py
```diff
@@ -0,0 +1,193 @@
...
+    def _CoreVerify(PK: BLSPubkey, message: bytes, signature: BLSSignature, DST: bytes) -> bool:
+        try:
+            signature_point = signature_to_G2(signature)
+            final_exponentiation = final_exponentiate(
+                pairing(
+                    signature_point,
+                    G1,
+                    final_exponentiate=False,
+                ) * pairing(
+                    hash_to_G2(message, DST),
```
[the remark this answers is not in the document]
CarlBeek: The problem here is that `PopProve` and `PopVerify` use different `DST`s when calling `_CoreSign` and `_CoreVerify`. One option is to have `DST` default to `None` and replace it with `cls.DST` in the function, but I thought that was ugly.

CarlBeek:
This PR is ready for merge! 🎉Thereafter we can merge `bls_standard` into `master` and cut a release.

PR #85, opened 2020-01-06
Update REAMDE for new BLS interface.

CarlBeek:
### What was wrong?

I forgot to update the README in #83.

### How was it fixed?

Updated the README.

#### Cute Animal Picture

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,59 +4,69 @@ Elliptic curve crypto in python including secp256k1 and alt_bn128
 
 [![GitHub](https://img.shields.io/github/license/mashape/apistatus.svg)](https://github.com/ethereum/py_ecc) [![Build Status](https://circleci.com/gh/ethereum/py_ecc.svg?style=shield)](https://circleci.com/gh/ethereum/py_ecc) [![PyPI version](https://badge.fury.io/py/py-ecc.svg)](https://badge.fury.io/py/py-ecc)
 
-
 ## Quickstart
+
 ```sh
 pip install py_ecc
 ```
 
 ## BLS Signatures
 
-```python
-from py_ecc import bls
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,59 +4,69 @@ Elliptic curve crypto in python including secp256k1 and alt_bn128
...
+
+```python
+from py_ecc.bls.G2ProofOfPossession import (
+    PrivToPub
+    Sign,
+    Verify,
+    Aggregate,
+    FastAggregateVerify,
+    AggregateVerify,
+)
```
hwwhww: It doesn't work since `G2ProofOfPossession` is a class.
We can either 
1. `from py_ecc.bls import G2ProofOfPossession as bls_pop` and then use `bls_pop.PrivToPub`, or
2. Move `G2Basic`, `G2MessageAugmentation`, `G2ProofOfPossession` to different files, and add APIs interface like:
```python
PrivToPub = G2ProofOfPossession.PrivToPub
Sign = G2ProofOfPossession.Sign
Verify = G2ProofOfPossession.Verify
Aggregate = G2ProofOfPossession.Aggregate
FastAggregateVerify = G2ProofOfPossession.FastAggregateVerify
AggregateVerify = G2ProofOfPossession.AggregateVerify
```

CarlBeek: Good catch! 👀 I think I'll go with option 1. Implementing now.

pipermerriam:
From what I understand this change to the API is a breaking one (it changes existing public APIs in a backwards incompatible way).  If so then we need to do a `2.0.0` release.  Can someone confirm/deny this?

hwwhww:
@pipermerriam Yes, that is correct.

PR #87, opened 2020-03-18
Switches hash-to-field to new v06 hash system

CarlBeek:
### What was wrong?

The BLS [hash-to-curve spec](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/) is being changed from v05 -> v06. BLS hash-to-curve  It is basically a change from repeated HKDF calls to direct SHA256 calls. This has the advantage of halving the number of SHA256 calls (and allows for other improvements for other curves in the hash-to-curve suite). See here for the full discussion: https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/issues/202

The changes made are summarised in PRs [212 (specs update)](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/pull/212), [214 (proof of concept implementation)](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/pull/214) & [217 (test vectors)](https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/pull/214).

### How was it fixed?

The new specs & tests were implemented.

**Please do not merge. This PR is for visibility for now until V06 is officially released.**

### To do:

- [x] Update to v06 when released (No changes expected)
- [x] Update hyperlinks to v06 spec when released

#### Cute Animal Picture

Baby Pangolin


py_ecc/bls/hash.py
```diff
@@ -30,8 +35,31 @@ def hkdf_expand(prk: Union[bytes, bytearray], info: Union[bytes, bytearray], len
...
+    return m.digest()
+
+
+def xor(a: bytes, b: bytes) -> bytes:
+    return bytes(_a ^ _b for _a, _b in zip(a, b))
+
+
+def expand_message_xmd(msg: bytes, DST: bytes, len_in_bytes: int) -> bytes:
+    ell = math.ceil(len_in_bytes / HASH_TO_FIELD_B_IN_BYTES)
+    DST_prime = len(DST).to_bytes(1, 'big') + DST
```
pipermerriam: I was curious about the performance characteristics of the various ways to convert a single integer value to a single byte.

The three that I evaluated were:

- `value.to_bytes(1, 'big')`
- `bytes([value])`
- `struct.pack('B', value)`

Both the `to_bytes` and the `bytes([value])` approach measured the same.  `struct.pack` however seems to be about 33% faster.

py_ecc/bls/hash.py
```diff
@@ -30,8 +35,31 @@ def hkdf_expand(prk: Union[bytes, bytearray], info: Union[bytes, bytearray], len
...
+
+
+def sha256(x: bytes) -> bytes:
+    m = _sha256()
+    m.update(x)
+    return m.digest()
+
+
+def xor(a: bytes, b: bytes) -> bytes:
+    return bytes(_a ^ _b for _a, _b in zip(a, b))
```
pipermerriam: Are you aware that this will truncate the extra characters if either `a, b` is longer than the other?

CarlBeek: I thought about adding an assert on the difference here, but I didn't want to add the overhead to every signature check. To be fair, the overhead is small in the context of the cost of hashing (dominated by the Fp^2 square root and the co-factor clearing). I only call this function for xor-ing `SHA256` digests. 

Options (not mutually exclusive) for handling this:
* Add in an `assert len(a) == len(b)`
* Rename the function to `_xor`
* zero-pad the shorter case (I am nervous of this behaviour as it is not obvious behaviour for an `xor` function)

py_ecc/bls/hash.py
CarlBeek: Awesome, thanks for checking this! 👍
I'll switch over to `struct.pack('B', value)`.

py_ecc/bls/hash.py
pipermerriam: I mostly just wanted to be sure you were aware of it.  I'll leave it to you to decide whether you think the extra safety is warranted as I'm not clear on how likely it would be for us to erroneously  pass in data that has a mismatched length.

Also, just a reminder that the stdlib has https://docs.python.org/3.5/library/itertools.html#itertools.zip_longest

py_ecc/bls/hash.py
pipermerriam: I just realized there is a much dumber solution that is around 6x faster.

```python
ALL_BYTES = tuple(bytes([i]) for i in range(256))
ord_to_byte = ALL_BYTES.index
```

This retains the property of throwing an error if the integer is out of range and changes it to be a constant speed lookup.

CarlBeek: It took me 6 days to come around to this proposal of yours, @pipermerriam. I was really torn between the beauty of its simplicity and its hackiness. Ultimately, I think the speed up is worth it here as the speed of this library is already a bottleneck for some projects. :)

py_ecc/bls/hash.py
```diff
@@ -30,8 +36,31 @@ def hkdf_expand(prk: Union[bytes, bytearray], info: Union[bytes, bytearray], len
...
+
+
+def expand_message_xmd(msg: bytes, DST: bytes, len_in_bytes: int) -> bytes:
```
hwwhww, line 56: What do you think about following the description of the input parameters of [the spec](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-06#section-5.3.1) by abstracting the hash function as one of the properties of the `BaseG2Ciphersuite` implementations?

e.g.,
```python
class G2Basic(BaseG2Ciphersuite):
    DST = b'BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_'
    xmd_hash_function = sha256
```

py_ecc/bls/hash.py
```diff
@@ -30,8 +36,31 @@ def hkdf_expand(prk: Union[bytes, bytearray], info: Union[bytes, bytearray], len
...
         okm.extend(previous)
 
     # Return first `length` bytes.
     return okm[:length]
+
+
+def sha256(x: bytes) -> bytes:
+    m = _sha256()
+    m.update(x)
+    return m.digest()
```
hwwhww, line 49: It seems we don't need to concatenate the `x` here, right?
Can we just use `return sha256(x).digest()`?

tests/bls/test_hash_to_curve.py
```diff
@@ -59,35 +60,30 @@ def test_iso_map_G2(iso_x, iso_y, iso_z, g2_x, g2_y):
     assert g2_y == result_y
 
 
+# Tests taken from: https://github.com/cfrg/draft-irtf-cfrg-hash-to-curve/blob/master/draft-irtf-cfrg-hash-to-curve.md#bls12381g2_xmdsha-256_sswu_ro_
 @pytest.mark.parametrize(
-    'msg,x,y',
+    'msg,x,y,H',
```
hwwhww, line 65: FYI (not strongly) we can also do
```python
@pytest.mark.parametrize('H', lambda x: sha256(x).digest())
@pytest.mark.parametrize(
    'msg,x,y',
    [
        ...
    ]
)
```

then `H` would be `lambda x: sha256(x).digest()` for all these test cases.

py_ecc/bls/hash.py
```diff
@@ -1,7 +1,14 @@
-import math
-import hashlib
 import hmac
-from typing import Union
+import math
+from typing import (
+    Union,
+)
+from hashlib import sha256
+from _hashlib import HASH  # type: ignore
```
CarlBeek, line 7: @hwwhww, I'd appreciate advice on typing here. `mypy` complains if this is not ignored and `flake8` doesn't like it if it is. I'm trying not to add additional mypy errors so #89 can be merged and mypy added to CI.

pipermerriam: You can make both of them ignore this by commenting like

```python
from _hashlib import HASH  # type: ignore  # noqa: <flake8-error-code>
```

Where `<flake8-error-code>` is something like `E501` (typically included in the error output from flake8

hwwhww: @CarlBeek sorry I missed you ping! 🙏
It seems fine to remove the comments: https://github.com/hwwhww/py_ecc/commit/39d8878471f987b4322dc209b56faf0b2043b0b2

I guess you were using different versions of flake8/mypy from the CI?

hwwhww:
@CarlBeek I pushed the fix into this branch, all the tests pass now.
Let me know if you think it's ready to merge. :)

CarlBeek:
Thanks for pushing the changes. I'll check my local flake8/mypy config. Good to merge!

PR #89, opened 2020-03-30
Fixes Typing errors for points at infinity (NoneTypes)

CarlBeek:
### What was wrong?

The point at infinity on any given curve is encoded as `None`. MyPy was not informed.

### How was it fixed?

I told MyPy.

#### Cute Animal Picture

image

### DIFF
--- a/py_ecc/bls12_381/bls12_381_curve.py
+++ b/py_ecc/bls12_381/bls12_381_curve.py
@@ -63,7 +63,7 @@ def is_inf(pt: GeneralPoint[Field]) -> bool:
 def is_on_curve(pt: Point2D[Field], b: Field) -> bool:
     if is_inf(pt):
         return True
-    x, y = pt  # type: ignore
+    x, y = pt
     return y**2 - x**3 == b
 
 
@@ -75,7 +75,7 @@ assert is_on_curve(G2, b2)
 def double(pt: Point2D[Field]) -> Point2D[Field]:
     if is_inf(pt):
         return pt
-    x, y = pt  # type: ignore
+    x, y = pt
     m = 3 * x**2 / (2 * y)
     newx = m**2 - 2 * x
     newy = -m * newx + m * x - y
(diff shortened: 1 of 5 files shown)


py_ecc/bls12_381/bls12_381_curve.py
```diff
@@ -63,7 +63,7 @@ def is_inf(pt: GeneralPoint[Field]) -> bool:
 def is_on_curve(pt: Point2D[Field], b: Field) -> bool:
     if is_inf(pt):
         return True
-    x, y = pt
+    x, y = pt  # type: ignore[misc]
```
CarlBeek: If anyone knows why mypy is so unhappy here, please let me know. The `if is_inf(pt)` check above checks for `None` so there is no way to reach line 66 with `pt = None`. MyPy doesn't realise this for some reason. :/

py_ecc/fields/field_elements.py
```diff
@@ -201,7 +201,7 @@ class FQP(object):
 
     def __init__(self,
                  coeffs: Sequence[IntOrFQ],
-                 modulus_coeffs: Sequence[IntOrFQ] = None) -> None:
+                 modulus_coeffs: Sequence[IntOrFQ] = []) -> None:
```
CarlBeek: This is to handle a separate error where `len(modulus_coeffs)` is called in line 208 below. This is not possible with `None` as it has no `__len__`. The reason for having a default argument here is for simple `type(self)` calls.

py_ecc/typing.py
```diff
@@ -64,8 +65,8 @@
     bls12_381_FQ2,
     bls12_381_FQ12,
 )
-Point2D = Tuple[Field, Field]
-Point3D = Tuple[Field, Field, Field]
+Point2D = Optional[Tuple[Field, Field]]  # Point at infinity is encoded as a None
```
CarlBeek: This is the meat of this PR and addresses 48 type errors.

Side note, should type checking be a part of CI on this repo? I feel like it should.

hwwhww: I opened issue #95 for this.

PR #91, opened 2020-05-04
BLS Hash to Curve v7

CarlBeek:
### What was wrong?
A new variant of H2C was released.

### How was it fixed?
Updated to v7.
There are 2 changes that have happened here. `sgn0` now uses a new method for calculating the sign and the domain length is appended to the `DST`.

I cannot find what is wrong here. It passes 1 of the 4 hash to curve test vectors. The problem is in [sgn0](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07#section-4.1), but I cannot find it. For example in the `a512_aaaaaaaaaaaaaaaaaaaaaaaa...` test, `Q0.x` is correct, but `Q0.y` is incorrect. I cannot trace where this comes from and it has been days... The function is so simple and I've implemented it line for line. **I give up.**


CarlBeek:
Closing as @hwwhww fixed my problem in #94. 🙏 The answer was a missing `property` decorator on `sgn0`.

PR #94, opened 2020-05-06
Replace #91: BLS Hash to Curve v7

hwwhww:
### What was wrong?
1. Rebased #91 
2. Make tests pass

### How was it fixed?
`FQ.sgn0`/`FQP.sgn0` should be property. I added annotation on it.

#### Cute Animal Picture
1559100129225_wombat_animal_cute


py_ecc/fields/optimized_field_elements.py
```diff
@@ -421,6 +422,23 @@ def __init__(self, coeffs: Sequence[IntOrFQ]) -> None:
...
+        Calculates the sign of a value.
+        sgn0(x) = 1 when x is 'negative'; otherwise, sg0(x) = 0
+
+        Note this is an optimized variant for m = 2
+
+        Defined here:
+        https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-07#section-4.1
+        """
+        x_0, x_1 = self.coeffs
+        sign_0 = x_0 % 2
```
CarlBeek, line 437: The modulus coefficients can be `FQ` elements or `int`s, so I think `mod_2(x)` should be implemented as a function with separate cases for int and FQ.

tests/bls/test_hash_to_curve.py
```diff
@@ -86,3 +86,27 @@ def test_hash_to_G2(msg, x, y, H):
...
+
+
+@pytest.mark.parametrize('value, expected', [
+    ([0x00, 0x00], 0),
+    ([0x10, 0x00], 0),
+    ([0x01, 0x00], 1),
+    ([0x01, 0x01], 1),
+    ([0x10, 0x10], 0),
+])
+def test_FQ2_sgn0(value, expected):
```
CarlBeek, line 110: This is a good idea. Where did you get the test vectors from? I think there should also be another test with the same parameters, but for the general case (`FQP`)

tests/bls/test_hash_to_curve.py
```diff
@@ -86,3 +86,27 @@ def test_hash_to_G2(msg, x, y, H):
...
     assert x == result_x
     assert y == result_y
+
+
+@pytest.mark.parametrize('value, expected', [
+    (0x00, 0),
+    (0x10, 0),
+    (0x01, 1),
+])
+def test_FQ_sgn0(value, expected):
```
CarlBeek, line 96: Should this test should also be applied to FQ2 and FQ12?

py_ecc/fields/optimized_field_elements.py
hwwhww: 👍    Added `mod_int` helper.

py_ecc/fields/optimized_field_elements.py
```diff
@@ -30,6 +30,15 @@
 IntOrFQ = Union[int, T_FQ]
 
 
+def mod_int(x: IntOrFQ, n: int) -> int:
+    if isinstance(x, int):
+        return x % n
+    elif isinstance(x, FQ):
+        return x.n % n
```
pipermerriam, line 37: Instead of special casing this, what do you think about implementing `FQ.__mod__`

https://docs.python.org/3/reference/datamodel.html#object.__mod__

py_ecc/fields/optimized_field_elements.py
```diff
@@ -428,6 +426,23 @@ def __init__(self, coeffs: Sequence[IntOrFQ]) -> None:
         self.mc_tuples = [(i, c) for i, c in enumerate(self.FQ2_MODULUS_COEFFS) if c]
         super().__init__(coeffs, self.FQ2_MODULUS_COEFFS)
 
+    @property
```
pipermerriam, line 429: Depending on how often these get accessed it might be worth leveraging `cached_property` for this and any other computed properties.

py_ecc/fields/optimized_field_elements.py
hwwhww: I was thinking about it, but there is a stub says `FQ.__mod__` is for `IntOrFQ`modulus:
```
def __mod__(self: T_FQ, other: IntOrFQ) -> T_FQ:
        raise NotImplementedError("Modulo Operation not yet supported by fields")
```
Having `mod_int` might be more explicit of the usage by its name.

CarlBeek:
IMO, this is ready for merge and release. @hwwhww, are you intending to include any other changes here?

hwwhww:
:shipit: !

PR #99, opened 2020-06-02
Update to draft-irtf-cfrg-hash-to-curve-08 test suite

hwwhww:
### What was wrong?

See if H2C v8 is compatible with H2C v7

### How was it fixed?

Only update the test vector per [ draft-irtf-cfrg-hash-to-curve-08](https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-08#appendix-H.10.1).


hwwhww:
replaced by #102

PR #100, opened 2020-06-08
Optimized BLS-12-381 Miller loop is now 33% faster

vbuterin:
Does the point multiplication step in the Miller loop in G2 rather than G12, increasing speed of computing the Miller loop by ~33% (0.2 sec -> 0.15 sec on my computer). Also includes a much faster twist routine.

### DIFF
--- a/py_ecc/optimized_bls12_381/optimized_pairing.py
+++ b/py_ecc/optimized_bls12_381/optimized_pairing.py
@@ -115,7 +115,7 @@ def miller_loop(Q: Optimized_Point3D[FQ2],
         return FQ12.one()
     cast_P = cast_point_to_fq12(P)
     twist_R = twist_Q = twist(Q)
-    R = Q  # type: Optimized_Point3D[FQ12]
+    R = Q  # type: Optimized_Point3D[FQ2]
     f_num, f_den = FQ12.one(), FQ12.one()
     # for i in range(log_ate_loop_count, -1, -1):
     for v in pseudo_binary_encoding[62::-1]:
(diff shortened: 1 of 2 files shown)


hwwhww:
I can only confirm that the output results remain the same against the Eth2 BLS test vectors.

py_ecc/optimized_bls12_381/optimized_pairing.py
```diff
@@ -109,30 +108,28 @@ def cast_point_to_fq12(pt: Optimized_Point3D[FQ]) -> Optimized_Point3D[FQ12]:
...
             f_den = f_den * _d
             R = add(R, Q)
-        elif v == -1:
-            nQ = neg(Q)
-            _n, _d = linefunc(R, nQ, P)
-            f_num = f_num * _n
-            f_den = f_den * _d
-            R = add(R, nQ)
+            twist_R = twist(R)
     # assert R == multiply(Q, ate_loop_count)
```
hwwhww, line 133: Is it okay to delete this and the following comments?

vbuterin: I suppose so?

captain-kark: Might be related to some of these [commented out tests](https://github.com/ethereum/py_ecc/blob/d21bbb11d4e7514956810e6ce1cd443a8e8ee817/tests/test_bn128_and_bls12_381.py#L328-L411), they're not using pytest correctly. Were they a part of the codebase back when runtime asserts *were* the test suite?

pipermerriam:
@vbuterin checking to be sure you have no futher changes for this PR

PR #103, opened 2020-09-23
Update to IETF BLS draft 04 and add some input validations

hwwhww:
### What was wrong?

Update to [IETF BLS draft 04](https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-04)

### How was it fixed?

1. `SK=0` (point at infinity PK) is disallowed in -04 spec. Update `KeyGen`, `KeyValidate` correspondingly.
2. Add the missing `KeyValidate` to `_CoreAggregateVerify`: It wasn't a bug because `_CoreAggregateVerify` itself has checked  `pubkey_to_G1`, but now it also needs to check the point at infinity PK.
3. Add input validations APIs `_is_valid_privkey`, `_is_valid_pubkey`, `_is_valid_message`, and `_is_valid_signature`. Reasons:
    - To check input types strictly since Python is a dynamically-typed language.
    - To follow the spec description more tightly.
    - As a workaround for PopVerify-able checks. e.g., ensure PK is not the point at infinity in `FastAggregateVerify`.

#### Cute Animal Picture
moose-70254_640

### DIFF
--- a/tests/bls/ciphersuites/test_g2_basic.py
+++ b/tests/bls/ciphersuites/test_g2_basic.py
@@ -2,15 +2,15 @@ import pytest
 from py_ecc.bls import G2Basic
 
 @pytest.mark.parametrize(
-    'SKs,messages,success',
+    'SKs,messages,result',
     [
-        (range(10), range(10), True),
-        (range(3), (b'42', b'69', b'42'), False),  # Test duplicate messages fail
+        (list(range(1, 11)), list(range(1, 11)), True),
+        (list(range(1, 4)), (b'42', b'69', b'42'), False),  # Test duplicate messages fail
     ]
 )
-def test_aggregate_verify(SKs, messages, success):
+def test_aggregate_verify(SKs, messages, result):
     PKs = [G2Basic.SkToPk(SK) for SK in SKs]
     messages = [bytes(msg) for msg in messages]
     signatures = [G2Basic.Sign(SK, msg) for SK, msg in zip(SKs, messages)]
     aggregate_signature = G2Basic.Aggregate(signatures)
-    assert G2Basic.AggregateVerify(PKs, messages, aggregate_signature) == success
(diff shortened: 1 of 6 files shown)


tests/bls/ciphersuites/test_g2_basic.py
```diff
@@ -1,16 +1,78 @@
...
+            G2Basic.Sign(privkey, message)
+
+
+
+@pytest.mark.parametrize(
+    'signatures, success',
+    [
+        ([G2Basic.Sign(1, b'helloworld')], True),
+        ([G2Basic.Sign(1, b'helloworld'), G2Basic.Sign(2, b'helloworld')], True),
+        ([Z2_SIGNATURE], True),
```
hwwhww, line 68: @CarlBeek could you help confirm 👀 that point at infinity signature is valid in `Aggregate`?

CarlBeek: I think this behaviour is undefined as the inputs to `Aggregate` are defined as "signature_1, ..., signature_n, octet strings output by either CoreSign or Aggregate."

CarlBeek (approved):
Aside from the validity of the point at infinity in the aggregate, this looks good to me. (To be clear, I both allowing it and disallowing it are fine with me as my reading of the specs has the behavior of `Aggregate` with the point at infinity as undefined.)

tests/bls/ciphersuites/test_g2_basic.py
hwwhww: I've opened an issue to get confirmation from IETF BLS standard maintainers: https://github.com/cfrg/draft-irtf-cfrg-bls-signature/issues/31

hwwhww: [The point at infinity is a valid aggregate signature because it is the sum of two valid signatures.](https://github.com/cfrg/draft-irtf-cfrg-bls-signature/issues/31#issuecomment-701157169)

PR #105, opened 2020-10-23
nit: rename py_ecc.bls.g2_primatives and add an alias

prestwich:
### What was wrong?

the `g2_primatives` module name was misspelled.

### How was it fixed?

Moved the file to a different name, and added an alias file to ensure imports under the old name work as expected.

#### Cute Animal Picture

Rattos


prestwich:
I've taken the liberty of moving the linting step ahead of the test runs in `tox.ini` so tests can fail faster.

prestwich:
closes #28

hwwhww (approved):
Thank you @prestwich!

My first thought was that we could replace `g2_primatives` with `g2_primitives` in the package internally, but then I saw that the status quo of this PR can ensure that both `g2_primatives` and `g2_primitives` are accessible, so either way looks fine to me.

pipermerriam (approved):

PR #106, opened 2020-11-05
Added bls import to init.py

vbuterin:
Fixes a bug where importing py_ecc does not make py_ecc.bls accessible.


pipermerriam:
I don't like releasing on Thursday/Fridays but I'll get a patch release out Monday.

PR #109, opened 2020-11-15
Add warning

hwwhww:
### What was wrong?

We should warn the users.

#### Cute Animal Picture

🕸️ 🕷️

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ pip install py_ecc
 
 ## BLS Signatures
 
+**Warning: It contains some experimental codes and has NOT been audited.**
+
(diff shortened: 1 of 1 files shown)


CarlBeek (commented):
I like this idea, but why specifically is the warning only in the BLS sections? Has the rest of py_ecc been audited?

Additionally, we should update line 3 to include a reference to the bls12-381 curve in addition to secp256k1 and alt_bn128.

README.md
```diff
@@ -1,6 +1,8 @@
 py_ecc
 ==========================
-Elliptic curve crypto in python including secp256k1 and alt_bn128
+Elliptic curve crypto in python including secp256k1, alt_bn128, and bls12_381.
+
+**Warning: It contains some experimental codes and has NOT been audited.**
```
pipermerriam, line 5:
```suggestion
> **Warning**: This library contains some experimental codes that have **NOT** been audited.
```

pipermerriam (approved):
Maybe good to be specific about what has/not been audited.  Has any of this been audited?

hwwhww:
> Has any of this been audited?

Afaik, I think no.

PR #110, opened 2020-11-16
Add strict deserialization format validation

### LINKED ISSUE #108: BLS point (de)compression issue
### What is wrong?

Follow up with #107, to see what's wrong with the (de)compression.

#### History
- https://github.com/ethereum/py_ecc/blob/master/py_ecc/bls/point_compression.py logic hasn't been updated for 1.5 years.
- The point compression format should have been following the old https://github.com/ethereum/eth2.0-specs/blob/v0.9.0/specs/bls_signature.md#point-representations doc, which was an explainer of Zcash docs: https://github.com/zcash/librustzcash/blob/6e0364cd42a2b3d2b958a54771ef51a8db79dd29/pairing/src/bls12_381/README.md#bls12-381-instantiation 
    - Respecting bit ordering, `z` is decomposed as `(c_flag, b_flag, a_flag, x)`.
    - The most significant bit (`c_flag`), when set, indicates that the point is in compressed form. Otherwise, the point is in uncompressed form.
    - The second-most significant bit (`b_flag`)indicates that the point is at infinity. If this bit is set, the remaining bits of the group element's encoding should be set to zero.
    - The third-most significant bit (`a_flag`) is set if (and only if) this point is in compressed form and it is not the point at infinity and its y-coordinate is the lexicographically largest of the two associated with the encoded x-coordinate.

#### The implementation and the given test case
- The flags of "normal" point at infinity PK: `110`
- The flags of the #107 case PK: `010`

The `decompress_G1` function: https://github.com/ethereum/py_ecc/blob/d17ee3cec483642e5a6bcd67c06d86eb6b3e7994/py_ecc/bls/point_compression.py#L55-L77

I think `decompress_G1` should have checked `c_flag` (`c_flag == 1` in `decompress_G1`). So do other (de)compress functions.

/cc @ChihChengLiang @CarlBeek could you 👀 sanity check it?

hwwhww:
### What was wrong?
Fix #108 

### How was it fixed?
Note: py_ecc doesn't provide APIs decompressed form deserialization, only compressed form deserialization.

1. Check if `c_flag is True`
2. If `b_flag` is True, check if `a_flag` is False and the value is 0.
3. For `decompress_G2`, check `c_flag2 == b_flag2 == a_flag2 == False`

#### Cute Animal Picture
🦩

### DIFF
--- a/py_ecc/bls/ciphersuites.py
+++ b/py_ecc/bls/ciphersuites.py
@@ -106,7 +106,7 @@ class BaseG2Ciphersuite(abc.ABC):
     def KeyValidate(PK: BLSPubkey) -> bool:
         try:
             pubkey_point = pubkey_to_G1(PK)
-        except ValidationError:
+        except Exception:
             return False
 
         if is_inf(pubkey_point):
@@ -114,7 +114,7 @@ class BaseG2Ciphersuite(abc.ABC):
 
         try:
             pubkey_subgroup_check(pubkey_point)
-        except ValidationError:
+        except Exception:
             return False
 
         return True
--- a/py_ecc/bls/ciphersuites.py
+++ b/py_ecc/bls/ciphersuites.py
@@ -106,7 +106,7 @@ class BaseG2Ciphersuite(abc.ABC):
     def KeyValidate(PK: BLSPubkey) -> bool:
         try:
             pubkey_point = pubkey_to_G1(PK)
-        except Exception:
+        except (ValidationError, ValueError, AssertionError):
             return False
 
         if is_inf(pubkey_point):
@@ -114,7 +114,7 @@ class BaseG2Ciphersuite(abc.ABC):
 
         try:
             pubkey_subgroup_check(pubkey_point)
-        except Exception:
+        except (ValidationError, ValueError, AssertionError):
             return False
 
         return True
--- a/py_ecc/bls/point_compression.py
+++ b/py_ecc/bls/point_compression.py
@@ -91,7 +91,7 @@ def decompress_G1(z: G1Compressed) -> G1Uncompressed:
     is_inf_pt = is_point_at_infinity(z)
 
     if b_flag != is_inf_pt:
-        raise ValueError(f"b_flag should be {int(is_inf_pt)}")
+        raise ValueError("b_flag should be %d" % int(is_inf_pt))
 
     if is_inf_pt:
         # 3 MSBs should be 110
@@ -224,7 +224,7 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
 
     # Validate z2 flags
     c_flag2, b_flag2, a_flag2 = get_flags(z2)
-    if not (c_flag2 == b_flag2 == a_flag2 and a_flag2 is False):
+    if not (c_flag2 is b_flag2 and b_flag2 is a_flag2 and a_flag2 is False):
         raise ValueError("a_flag2, b_flag2, and c_flag2 should always set to 0")
 
     return (x, y, FQ2([1, 0]))
--- a/py_ecc/bls/point_compression.py
+++ b/py_ecc/bls/point_compression.py
@@ -45,6 +45,10 @@ def get_flags(z: int) -> Tuple[bool, bool, bool]:
 
 
 def is_point_at_infinity(z1: int, z2: Optional[int] = None) -> bool:
+    """
+    If z2 is None, the given z1 is a G1 point.
+    Else, (z1, z2) is a G2 point.
+    """
     return (z1 % POW_2_381 == 0) and (
         z2 is None or (z2 is not None and z2 == 0)
     )
--- a/py_ecc/bls/point_compression.py
+++ b/py_ecc/bls/point_compression.py
@@ -194,6 +194,12 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
     # Else, not point at infinity
     # 3 MSBs should be 100 or 101
     x1 = z1 % POW_2_381
+
+    # Validate z2 flags
+    c_flag2, b_flag2, a_flag2 = get_flags(z2)
+    if not (c_flag2 is b_flag2 and b_flag2 is a_flag2 and a_flag2 is False):
+        raise ValueError("a_flag2, b_flag2, and c_flag2 should always set to 0")
+
     x2 = z2
     # x1 is the imaginary part, x2 is the real part
     x = FQ2([x2, x1])
@@ -214,10 +220,4 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
         raise ValueError(
             "The given point is not on the twisted curve over FQ**2"
         )
-
-    # Validate z2 flags
-    c_flag2, b_flag2, a_flag2 = get_flags(z2)
-    if not (c_flag2 is b_flag2 and b_flag2 is a_flag2 and a_flag2 is False):
-        raise ValueError("a_flag2, b_flag2, and c_flag2 should always set to 0")
-
     return (x, y, FQ2([1, 0]))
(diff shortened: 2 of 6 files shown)


py_ecc/bls/point_compression.py
```diff
@@ -30,6 +31,52 @@
...
+        z2 is None or (z2 is not None and z2 == 0)
+    )
+
+
+def validate_point_at_infinity(z1: int, a_flag: bool, z2: Optional[int] = None) -> None:
+    """
+    If z2 is None, the given z1 is a G1 point.
+    Else, (z1, z2) is a G2 point.
+    """
+    if a_flag == 0:
```
ChihChengLiang, line 73: since a_flag is bool
```suggestion
    if not a_flag:
```

py_ecc/bls/point_compression.py
```diff
@@ -161,4 +230,12 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
         raise ValueError(
             "The given point is not on the twisted curve over FQ**2"
         )
+
+    # Validate z2 flags
+    c_flag2 = get_c_flag(z2)
+    b_flag2 = get_b_flag(z2)
+    a_flag2 = get_a_flag(z2)
```
pipermerriam, line 237: You might wrap this up into something like `c_flag2, b_flag2, a_flag2 = get_flags(z2)`

py_ecc/bls/point_compression.py
```diff
@@ -161,4 +230,12 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
...
+    if not (c_flag2 == b_flag2 == a_flag2 and c_flag2 is False):
```
pipermerriam, line 238: Assuming you're going for "strict" checking (that all of these values are boolean `False` values) then you would need to do `c_flag2 is b_flag2 and b_flag2 is a_flag2 and a_flag2 is False`

pipermerriam (approved):

py_ecc/bls/point_compression.py
```diff
@@ -56,10 +103,23 @@ def decompress_G1(z: G1Compressed) -> G1Uncompressed:
     """
     Recovers x and y coordinates from the compressed point.
     """
-    # b_flag == 1 indicates the infinity point
-    b_flag = (z % POW_2_383) // POW_2_382
-    if b_flag == 1:
+    c_flag = get_c_flag(z)
+    b_flag = get_b_flag(z)
+    a_flag = get_a_flag(z)
```
ChihChengLiang, line 108: What do you think if we can get more performance and readability by using bit operations?

```python
c_flag , b_flag, a_flag = get_flags(z)
def get_flags(z):
    a_flag = bool((z >> 380) & 1)
    b_flag = bool((z >> 381) & 1)
    c_flag = bool((z >> 382) & 1)
    return c_flag , b_flag, a_flag
```
```python
>>> import random
>>> for _ in range(10):
...     z = random.getrandbits(384)
...     print(get_flags(z))
... 
(True, False, True)
(True, True, True)
(True, True, False)
(False, True, False)
(False, False, False)
(True, True, True)
(True, True, False)
(True, True, False)
(True, False, True)
(False, True, True)
```

py_ecc/bls/point_compression.py
```diff
@@ -56,10 +103,23 @@ def decompress_G1(z: G1Compressed) -> G1Uncompressed:
...
+    if not c_flag:
+        raise ValueError("c_flag should be 1")
+
+    # b_flag == 1 indicates the point at infinity
+    if b_flag:
+        validate_point_at_infinity(z1=z, a_flag=a_flag)
         return Z1
+    else:
+        if is_point_at_infinity(z):
+            raise ValueError("b_flag should be 1")
```
ChihChengLiang, line 120: What do you think
```suggestion
            raise ValueError("a point at infinity should have b_flag 1")
```

py_ecc/bls/ciphersuites.py
```diff
@@ -106,15 +106,15 @@ def KeyGen(cls, IKM: bytes, key_info: bytes = b'') -> int:
...
+        except Exception:
             return False
 
         if is_inf(pubkey_point):
             return False
 
         try:
             pubkey_subgroup_check(pubkey_point)
-        except ValidationError:
+        except Exception:
```
ChihChengLiang, line 117: `Exception` is kind of broad, should we catch `ValidationError` and `ValueError`?

py_ecc/bls/point_compression.py
ChihChengLiang, line 238: what do you think
```suggestion
    if not (c_flag2 == b_flag2 == a_flag2 == False):
```

tests/bls/test_g2_core.py
```diff
@@ -16,7 +16,7 @@
     'pubkey,success',
     [
         (G2Basic.SkToPk(42), True),
-        (b'11' * 48, False),
+        (b'\x11' * 48, False),
```
ChihChengLiang, line 19: nice catch!

ChihChengLiang (approved):
Looks good to me. Added some nitpicks.

py_ecc/bls/point_compression.py
hwwhww: `a_flag2 == False` would make linter unhappy. 😅 
I applied Piper's suggestion!

py_ecc/bls/point_compression.py
```diff
@@ -30,6 +30,30 @@
...
+
+
+def is_point_at_infinity(z1: int, z2: Optional[int] = None) -> bool:
+    """
+    If z2 is None, the given z1 is a G1 point.
+    Else, (z1, z2) is a G2 point.
+    """
+    return (z1 % POW_2_381 == 0) and (
+        z2 is None or (z2 is not None and z2 == 0)
+    )
```
ChihChengLiang, line 54: Looks like equivalent to
```suggestion
    return (z1 % POW_2_381 == 0) and (z2 is None or z2 == 0)
```

py_ecc/bls/point_compression.py
```diff
@@ -56,10 +80,27 @@ def decompress_G1(z: G1Compressed) -> G1Uncompressed:
...
-    b_flag = (z % POW_2_383) // POW_2_382
-    if b_flag == 1:
+    c_flag, b_flag, a_flag = get_flags(z)
+
+    # c_flag == 1 indicates the compressed form
+    # MSB should be 1
+    if not c_flag:
+        raise ValueError("c_flag should be 1")
+
+    # b_flag == 1 indicates the point at infinity
```
ChihChengLiang, line 90:
```suggestion
```

py_ecc/bls/point_compression.py
```diff
@@ -136,12 +176,30 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
...
     z1, z2 = p
+    c_flag1, b_flag1, a_flag1 = get_flags(z1)
+
+    # c_flag == 1 indicates the compressed form
+    if not c_flag1:
+        raise ValueError("c_flag should be 1")
+
+    # MSB should be 1
+    if not c_flag1:
+        raise ValueError("c_flag should be 1")
```
ChihChengLiang, line 187: Duplication
```suggestion
```

py_ecc/bls/point_compression.py
```diff
@@ -136,12 +176,30 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
...
 
-    # b_flag == 1 indicates the infinity point
-    b_flag1 = (z1 % POW_2_383) // POW_2_382
-    if b_flag1 == 1:
+    # b_flag == 1 indicates the point at infinity
```
ChihChengLiang, line 189:
```suggestion
```

ChihChengLiang (approved):
Found some duplications, other looks good to me.

hwwhww:
Thanks for the reviews! :)

PR #116, opened 2021-02-08
Extract `subgroup_check` from `signature_to_G2`

hwwhww:
### What was wrong?
`signature_to_G2` should be only for deserialization and decompression. 

### How was it fixed?
Extract `subgroup_check(signature_point)` from `signature_to_G2`.

#### Cute Animal Picture

🦐


CarlBeek (approved):
Definitely cleans things up. 👍

Don't know why CI is not running, but tests pass locally.

hwwhww:
I just summoned CI to run. 🧙 Merging now!

PR #117, opened 2021-02-21
Add G1 and G2 point value check

hwwhww:
### What was wrong?

Ensure the value is less than the field modulus.

### How was it fixed?

Add more checks.

#### Cute Animal Picture

🦍


CarlBeek (approved):
LGTM!

PR #121, opened 2021-04-07
Fix the x1 point check

hwwhww:
### What was wrong?
A patch for #117
`x1` size check was missing.

Credit to @asanso for finding and reporting it! 🙏

### How was it fixed?

Add `if x1 >= q` check and test case.

#### Cute Animal Picture

🦑

### DIFF
--- a/py_ecc/bls/point_compression.py
+++ b/py_ecc/bls/point_compression.py
@@ -196,6 +196,8 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
     # Else, not point at infinity
     # 3 MSBs should be 100 or 101
     x1 = z1 % POW_2_381
+    if x1 >= q:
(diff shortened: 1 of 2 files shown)


py_ecc/bls/point_compression.py
```diff
@@ -196,6 +196,8 @@ def decompress_G2(p: G2Compressed) -> G2Uncompressed:
     # Else, not point at infinity
     # 3 MSBs should be 100 or 101
     x1 = z1 % POW_2_381
+    if x1 >= q:
```
hwwhww, line 199: Add comment
```suggestion
    # Ensure that x1 is less than the field modulus.
    if x1 >= q:
```

CarlBeek (approved):
Glad we're patching this hole completely now.

PR #123, opened 2021-11-18
Upgrade eth-typing and eth-utils

kclowes:
### What was wrong?
I dropped python 3.5 support from eth-typing, and eth-utils, so had to do a major version bump of those. py-evm depends on py_ecc, so had to resolve dependency conflicts. 

### How was it fixed?
Bumped dependency versions of `eth-typing` and `eth-utils`.

Had to drop python3.5 support here as well, and added in python 3.9 and 3.10 support. I can add the python3.9 and 3.10 support in another PR if anyone feels strongly.

#### Cute Animal Picture


carver (approved):
LGTM. Are these changes making it into the template?

kclowes:
Good call! I did not remember that we had a template. Will add!

PR #112, opened 2020-11-26
Setup towncrier to generate release notes

tmckenzie51:
### What was wrong?
Release notes process is not automated.

### How was it fixed?
Towncrier is setup in order to generate the release notes. Similar changes made as shown in the template https://github.com/ethereum/ethereum-python-project-template/pull/28#issue-302025093 

#### Cute Animal Picture

put a cute animal picture link inside the parentheses/GettyImages-634869043-58a6e83f5f9b58a3c918ca12.jpg)

PR #130, opened 2023-10-29
Upgrade template

pacrob:
### What was wrong?

Needs some general upkeep.

### How was it fixed?

Lots of updates from the python project template, notably:
 - drop python 3.6 and 3.7 support
 - add python 3.11 support
 - move to using pre-commit for linting
 - use towncrier for handling newsfragments

Closes #112 

#### Cute Animal Picture

image


py_ecc/bls/g2_primatives.py
```diff
@@ -1,4 +1,3 @@
 # flake8: noqa
 # `g2_primatives.py` was a typo of `g2_primitives.py`
 # Make an alias for backward compatibility
-from py_ecc.bls.g2_primitives import *
```
kclowes, line 4: If we're deleting this line/making a breaking change to the library, I think this file can go?

py_ecc/bls/point_compression.py
```diff
@@ -121,16 +122,19 @@ def decompress_G1(z: G1Compressed) -> G1Uncompressed:
...
+    Given value=``x``, returns the value ``y`` such that ``y**2 % q == x``,
     and None if this is not possible. In cases where there are two solutions,
     the value with higher imaginary component is favored;
     if both solutions have equal imaginary component the value with higher real
     component is favored.
     """
     candidate_squareroot = value ** ((FQ2_ORDER + 8) // 16)
-    check = candidate_squareroot ** 2 / value
+    check = candidate_squareroot**2 / value
     if check in EIGTH_ROOTS_OF_UNITY[::2]:
```
kclowes, line 133: This can be put in a separate PR/issue, but I believe this should be: `EIGHTH_ROOTS_OF_UNITY`

py_ecc/fields/field_elements.py
```diff
@@ -49,8 +50,9 @@ def __init__(self: T_FQ, val: IntOrFQ) -> None:
             self.n = val % self.field_modulus
         else:
             raise TypeError(
-                "Expected an int or FQ object, but got object of type {}"
-                .format(type(val))
+                "Expected an int or FQ object, but got object of type {}".format(
```
kclowes, line 53: Might be nice to make these f-strings at some point?

py_ecc/fields/optimized_field_elements.py
```diff
@@ -1,20 +1,15 @@
-try:
-    # Python 3.8
-    from functools import cached_property  # type: ignore
-except (ImportError, SyntaxError):
-    # Python 3 to 3.7
-    from cached_property import cached_property
```
kclowes, line 6: Amazing!

tests/bls/test_g2_core.py
```diff
@@ -1,52 +1,52 @@
 import pytest
 
-from py_ecc.optimized_bls12_381 import (
-    G1,
-    G2,
-    multiply,
+from py_ecc.bls import (
+    G2Basic,
 )
 from py_ecc.bls.g2_primatives import (
```
kclowes, line 6: Should this be `g2_primitives`?

tests/bls/test_g2_primatives.py
```diff
@@ -1,32 +1,31 @@
 import pytest
 
+from py_ecc.bls.g2_primatives import (
```
kclowes, line 3: same here: `primatives` -> `primitives`?

tests/test_backward_compatibility.py
```diff
@@ -1,39 +1,10 @@
...
-        FQ12,
-    )
-    from py_ecc.optimized_bls12_381 import (
-        field_modulus,
-        FQ,
-        FQP,
-        FQ2,
-        FQ12,
-    )
+    pass
```
kclowes, line 2: Do these get tested elsewhere?

kclowes (approved):
Thanks for getting this in! I had a couple misc questions/comments throughout, but nothing big. One thing I thought of is: while we're at it, maybe we should test against Python 3.12 too now that it's out. If it's going to be a whole thing :tm:, then we can exclude it for now, but if it's as simple as adding the tests to CI, maybe we can toss it in too? Or it can wait for the next PR as well. Whatever you think is best. 🎉

pacrob:
> test against Python 3.12

I'm planning to do that as a separate PR from the template and see how just cherry-picking the same commit works. And I need to start further up the dep tree, probably with `eth-utils`.

py_ecc/bls/g2_primatives.py
pacrob: removed & updated refs

tests/bls/test_g2_core.py
pacrob: updated

tests/bls/test_g2_primatives.py
pacrob: updated

pacrob: updated

py_ecc/bls/point_compression.py
pacrob: fine in this PR I think, updated everywhere

py_ecc/fields/field_elements.py
pacrob: Got'em, plus some %-strings.

kclowes:
Cool, that makes sense!

PR #143, opened 2024-07-30
Update typing

### LINKED ISSUE #131: Update type hinting
Types have been added throughout the lib, but against an outdated version of `mypy` that won't run with recent versions of Python. `mypy` is being added as part of `pre-commit`, but set to ignore all files. As typing is updated, they can be added to be checked.

pacrob:
### What was wrong?

mypy checks had been turned off.

Closes #131 

### How was it fixed?

Move mypy checks to run locally, fix errors

The largest changes are within `field_elements` and `optimized_field_elements`, where the `IntOrFQ` type changed from using a generic to a regular type (`FQ` instead of `T_FQ`). I was unable to resolve the use of the generic `T_FQ` within the `FQP` class when it was bound to the `FQ` class, and since it was only used within functions, changing to `FQ` seems to work fine.

The change had some downstream effects, which were resolved either by `int`ing `IntOrFQ` variables (`FQ` defines `__int__` so that this is handled correctly, or by defining the `__lt__`. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/py_ecc/fields/field_elements.py
+++ b/py_ecc/fields/field_elements.py
@@ -1,3 +1,6 @@
+from functools import (
+    total_ordering,
+)
 from typing import (  # noqa: F401
     TYPE_CHECKING,
     Any,
@@ -30,9 +33,10 @@ T_FQ = TypeVar("T_FQ", bound="FQ")
 T_FQP = TypeVar("T_FQP", bound="FQP")
 T_FQ2 = TypeVar("T_FQ2", bound="FQ2")
 T_FQ12 = TypeVar("T_FQ12", bound="FQ12")
-IntOrFQ = Union[int, T_FQ]
+IntOrFQ = Union[int, "FQ"]
 
 
+@total_ordering
 class FQ:
     """
     A class for field elements in FQ. Wrap a number in this class,
@@ -42,7 +46,7 @@ class FQ:
     n: int
     field_modulus: int
 
-    def __init__(self: T_FQ, val: IntOrFQ[T_FQ]) -> None:
+    def __init__(self: T_FQ, val: IntOrFQ) -> None:
         if not hasattr(self, "field_modulus"):
             raise AttributeError("Field Modulus hasn't been specified")
 
@@ -55,7 +59,7 @@ class FQ:
                 f"Expected an int or FQ object, but got object of type {type(val)}"
             )
 
-    def __add__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __add__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         if isinstance(other, FQ):
             on = other.n
         elif isinstance(other, int):
@@ -67,7 +71,7 @@ class FQ:
 
         return type(self)((self.n + on) % self.field_modulus)
 
-    def __mul__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __mul__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         if isinstance(other, FQ):
             on = other.n
         elif isinstance(other, int):
@@ -79,13 +83,13 @@ class FQ:
 
         return type(self)((self.n * on) % self.field_modulus)
 
-    def __rmul__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __rmul__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         return self * other
 
-    def __radd__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __radd__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         return self + other
 
-    def __rsub__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __rsub__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         if isinstance(other, FQ):
             on = other.n
         elif isinstance(other, int):
@@ -97,7 +101,7 @@ class FQ:
 
         return type(self)((on - self.n) % self.field_modulus)
 
-    def __sub__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __sub__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         if isinstance(other, FQ):
             on = other.n
         elif isinstance(other, int):
@@ -109,7 +113,7 @@ class FQ:
 
         return type(self)((self.n - on) % self.field_modulus)
 
-    def __div__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __div__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         if isinstance(other, FQ):
             on = other.n
         elif isinstance(other, int):
@@ -123,10 +127,10 @@ class FQ:
             self.n * prime_field_inv(on, self.field_modulus) % self.field_modulus
         )
 
-    def __truediv__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __truediv__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         return self.__div__(other)
 
-    def __rdiv__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __rdiv__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         if isinstance(other, FQ):
             on = other.n
         elif isinstance(other, int):
@@ -140,7 +144,7 @@ class FQ:
             prime_field_inv(self.n, self.field_modulus) * on % self.field_modulus
         )
 
-    def __rtruediv__(self: T_FQ, other: IntOrFQ[T_FQ]) -> T_FQ:
+    def __rtruediv__(self: T_FQ, other: IntOrFQ) -> T_FQ:
         return self.__rdiv__(other)
 
     def __pow__(self: T_FQ, other: int) -> T_FQ:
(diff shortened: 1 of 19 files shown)


py_ecc/fields/field_elements.py
```diff
@@ -254,8 +270,10 @@ def __rmul__(self: T_FQP, other: Union[int, T_FQP]) -> T_FQP:
         return self * other
 
     def __div__(self: T_FQP, other: Union[int, T_FQP]) -> T_FQP:
-        if isinstance(other, int_types_or_FQ):
-            return type(self)([c / other for c in self.coeffs])
+        if isinstance(other, int):
+            return type(self)(
+                [c / other if isinstance(c, FQ) else c // other for c in self.coeffs]
```
pacrob, line 275: The return type requires that the list be `int`s. `/` is defined for `FQ` such that it will be an int, but in the case that `c` is an int, `//` is needed to guarantee the `int`. The other option I thought would be to make the `int` an `FQ` and then just use `/`.

py_ecc/fields/field_elements.py
```diff
@@ -28,20 +32,21 @@
 T_FQP = TypeVar("T_FQP", bound="FQP")
 T_FQ2 = TypeVar("T_FQ2", bound="FQ2")
 T_FQ12 = TypeVar("T_FQ12", bound="FQ12")
-IntOrFQ = Union[int, T_FQ]
+IntOrFQ = Union[int, "FQ"]
```
reedsa, line 35: Not quite sure if there's benefit to the type checker, does this work the same if using `Literal` here? Just curious for my learning.
```suggestion
IntOrFQ = Union[int, Literal["FQ"]]
```

py_ecc/fields/field_elements.py
```diff
@@ -173,6 +178,17 @@ def __repr__(self: T_FQ) -> str:
     def __int__(self: T_FQ) -> int:
         return self.n
 
+    def __lt__(self: T_FQ, other: IntOrFQ) -> bool:
```
reedsa, line 181: Should `other` be `Any` here to accommodate the `else`?
```suggestion
    def __lt__(self: T_FQ, other: Any) -> bool:
```

py_ecc/fields/field_elements.py
```diff
@@ -254,8 +270,10 @@ def __rmul__(self: T_FQP, other: Union[int, T_FQP]) -> T_FQP:
         return self * other
 
     def __div__(self: T_FQP, other: Union[int, T_FQP]) -> T_FQP:
-        if isinstance(other, int_types_or_FQ):
-            return type(self)([c / other for c in self.coeffs])
+        if isinstance(other, int):
```
reedsa, line 273: Is it possible that `other` should include the `FQ` type in the Union? Not sure if this function was ever taking an argument of that type, do you have more context there?

py_ecc/fields/field_elements.py
```diff
@@ -349,11 +366,11 @@ class FQ2(FQP):
...
     """
 
-    degree = 2
-    FQ2_MODULUS_COEFFS = None  # type: FQ2_modulus_coeffs_type
+    degree: int = 2
+    FQ2_MODULUS_COEFFS: "FQ2_modulus_coeffs_type"
 
     def __init__(self, coeffs: Sequence[IntOrFQ]) -> None:
-        if self.FQ2_MODULUS_COEFFS is None:
+        if not hasattr(self, "FQ2_MODULUS_COEFFS"):
```
reedsa, line 373: Should this allow a value of `None` if the attribute is present? I think the `is None` check is still needed.

Another option is to first assign a variable like `x = self.get("FQ2_MODULUS_COEFFS")` and then use the same logic as before.

py_ecc/fields/field_elements.py
```diff
@@ -364,11 +381,11 @@ class FQ12(FQP):
...
     """
 
-    degree = 12
-    FQ12_MODULUS_COEFFS = None  # type: FQ12_modulus_coeffs_type
+    degree: int = 12
+    FQ12_MODULUS_COEFFS: "FQ12_modulus_coeffs_type"
 
     def __init__(self, coeffs: Sequence[IntOrFQ]) -> None:
-        if self.FQ12_MODULUS_COEFFS is None:
+        if not hasattr(self, "FQ12_MODULUS_COEFFS"):
```
reedsa, line 388: Same check here as above.

py_ecc/fields/optimized_field_elements.py
```diff
@@ -30,7 +32,7 @@
 T_FQP = TypeVar("T_FQP", bound="FQP")
 T_FQ2 = TypeVar("T_FQ2", bound="FQ2")
 T_FQ12 = TypeVar("T_FQ12", bound="FQ12")
-IntOrFQ = Union[int, T_FQ]
+IntOrFQ = Union[int, "FQ"]
```
reedsa, line 35: Another `Literal` needed here?

py_ecc/fields/optimized_field_elements.py
```diff
@@ -39,20 +41,21 @@ def mod_int(x: IntOrFQ, n: int) -> int:
...
     """
 
-    n = None  # type: int
-    field_modulus = None  # type: int
+    n: int
+    field_modulus: int
 
     def __init__(self: T_FQ, val: IntOrFQ) -> None:
-        if self.field_modulus is None:
+        if not hasattr(self, "field_modulus"):
```
reedsa, line 58: Does this still require the `None` logic like the others?

py_ecc/fields/optimized_field_elements.py
```diff
@@ -187,6 +190,17 @@ def __repr__(self: T_FQ) -> str:
     def __int__(self: T_FQ) -> int:
         return self.n
 
+    def __lt__(self: T_FQ, other: IntOrFQ) -> bool:
```
reedsa, line 193:
```suggestion
    def __lt__(self: T_FQ, other: Any) -> bool:
```

py_ecc/fields/optimized_field_elements.py
```diff
@@ -215,13 +229,13 @@ class FQP:
...
-    field_modulus: Union[int, None] = None
-    mc_tuples: Union[List[Tuple[int, int]], None] = None
+    field_modulus: int
+    mc_tuples: List[Tuple[int, int]]
 
     def __init__(
         self, coeffs: Sequence[IntOrFQ], modulus_coeffs: Sequence[IntOrFQ] = ()
     ) -> None:
-        if self.field_modulus is None:
+        if not hasattr(self, "field_modulus"):
```
reedsa, line 238: `None` check here as well?

py_ecc/fields/optimized_field_elements.py
```diff
@@ -408,11 +422,11 @@ class FQ2(FQP):
...
     """
 
-    degree = 2
-    FQ2_MODULUS_COEFFS = None  # type: FQ2_modulus_coeffs_type
+    degree: int = 2
+    FQ2_MODULUS_COEFFS: "FQ2_modulus_coeffs_type"
 
     def __init__(self, coeffs: Sequence[IntOrFQ]) -> None:
-        if self.FQ2_MODULUS_COEFFS is None:
+        if not hasattr(self, "FQ2_MODULUS_COEFFS"):
```
reedsa, line 429: `None` check?

py_ecc/fields/optimized_field_elements.py
```diff
@@ -441,11 +455,11 @@ class FQ12(FQP):
...
     """
 
-    degree = 12
-    FQ12_MODULUS_COEFFS = None  # type: FQ12_modulus_coeffs_type
+    degree: int = 12
+    FQ12_MODULUS_COEFFS: "FQ12_modulus_coeffs_type"
 
     def __init__(self, coeffs: Sequence[IntOrFQ]) -> None:
-        if self.FQ12_MODULUS_COEFFS is None:
+        if not hasattr(self, "FQ12_MODULUS_COEFFS"):
```
reedsa, line 462: `None` check?

reedsa (commented):
Thanks for working through all of these! Had a few questions to solidify my understanding, not totally sure about the logic in some cases.

py_ecc/fields/field_elements.py
pacrob: It's not a `Literal`, it's how to get mypy to approve of the use of a class or type before it's actually defined.

py_ecc/fields/field_elements.py
pacrob: The only dunder methods that complained about not being `Any` are `__eq__` and `__ne__`. All the typed examples I can find of comparison dunders have `other` being of the same type as `self`. This seems to be a special case where both `FQ` and `int` are kind of considered the same.

Here's a similar example from `pycryptodome` - https://github.com/Legrandin/pycryptodome/blob/13a8461e068ebca99d21e0f2f71830504b25ac7e/lib/Crypto/Math/_IntegerBase.pyi#L17

py_ecc/fields/field_elements.py
pacrob: You're right, there's a conflict with `other` type, `if` options, and the error message. Will work more on this.

py_ecc/fields/field_elements.py
pacrob: This changed along with changing how the attributes of these classes were defined. Previously, it was:

`FQ12_MODULUS_COEFFS = None  # type: FQ12_modulus_coeffs_type`

This then makes it either the `FQ12_modulus_coeffs_type` or `None`, which then required a ton of addition validation through implementations to make sure that it's not `None` whenever it's used. If it's just declared without assignment, then it's there, and typed, ready for subclasses to assign a value to it without requiring the extra `if not None` nonsense.

py_ecc/fields/field_elements.py
pacrob: Digging back through blames, we can see the application of types here:
https://github.com/ethereum/py_ecc/commit/0309225c56acdb081f535ff6353fff2cad4f8329#diff-93c798286fdea4df041291e658b99539423804a846045a425261d05b2e0e6691

My best guess is that the dropping of the `FQ` option was a typo. We can see `__mul__`, `__rmul__`, `__div__` and `__truediv__` all got types updated to use the generics, but only `__mul__` kept `int, T_FQ`, `T_FQP`, with no logic change in any of the functions. This doesn't make sense, at least for `__rmul__` because it's passing `other` over to `__mul__` which still says it accepts `T_FQ`.

So yeah, I think `FQ` was dropped in error. Adding it back.

fselmo (approved):
lgtm!

PR #144, opened 2024-08-22
Lazy-load submodules

pacrob:
### What was wrong?

Import of `py_ecc` in other libs takes a long time.

Related to Issue #135 

### How was it fixed?

Lazy-load all submodules.
Added `import sys as _sys` to clarify intent on calling `dir(py_ecc)`.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/135.performance.rst
+++ b/newsfragments/135.performance.rst
@@ -1,0 +1,1 @@
+Lazy-load submodules to reduce initial import time
--- a/py_ecc/__init__.py
+++ b/py_ecc/__init__.py
@@ -1,17 +1,40 @@
+import importlib
 from importlib.metadata import (
     version as __version,
 )
-import sys
-
-from py_ecc import (
-    bls,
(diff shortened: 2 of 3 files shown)


reedsa (approved):
This is cool, lgtm!

fselmo (approved):
This can potentially hide some runtime issues when importing the modules. I think in  this should be pretty well covered by our test suite just running, but as a sanity check we should probably have a test that imports all of these to make absolutely sure there are no runtime errors. Can probably just use something like:

```py
importlib.importmodule(f"py_ecc.{submod}") for submod in lazy_load_submodules
```

Looks good to me otherwise.

fselmo:
@pacrob I think it could also be helpful to add a quick reason for the lazy-loading to the git blame (commit message) for context.

fselmo:
lgtm 👌🏼

PR #145, opened 2024-08-28
set ecdsa_raw_recover to accept v values of 27 or 28

pacrob:
### What was wrong?

Accepted values for `v` in `ecdsa_raw_recover` appear incorrect, and don't match error message.
Originally noted in `eth-keys` - https://github.com/ethereum/eth-keys/issues/67 - but same bug here.

### How was it fixed?

Set acceptable `v` values to be 27 or 28. See https://github.com/ethereum/eth-keys/pull/100 for detailed explanation. Though there, because the value passed was 0 or 1, it was noted that the value is aka `y-parity`. In this case, because it is passed as 27 or 28 and the doc string describes it as `v`, I've left the error message just referencing `v`.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/newsfragments/145.breaking.rst
+++ b/newsfragments/145.breaking.rst
@@ -1,0 +1,1 @@
+Set ``ecdsa_raw_recover`` to only accept ``v`` values of 27 or 28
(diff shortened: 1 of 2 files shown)


kclowes (approved):
LGTM!

fselmo (commented):
👍🏼

PR #146, opened 2024-09-06
Implement hash to curve functions for G1 curve

petertdavies:
### What was wrong?

Hash to curve functions for the G1 curve were not included.

### How was it fixed?

Added hash to curve functions for the G1 curve. These are needed for [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537).

### Todo:

#### Cute Animal Picture


py_ecc/bls/hash_to_curve.py
```diff
@@ -94,3 +98,27 @@ def clear_cofactor_G2(p: G2Uncompressed) -> G2Uncompressed:
...
+
+    First, convert FQ1 point to a point on the 3-Isogeny curve.
+    SWU Map: https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-09#section-6.6.3
+
+    Second, map 3-Isogeny curve to BLS12-381-G2 curve.
+    3-Isogeny Map:
+    https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-09#name-11-isogeny-map-for-bls12-38
+    """
+    (x, y, z) = optimized_swu_G1(u)
+    return iso_map_G1(x, y, z)
```
pacrob, line 115: @petertdavies I'm seeing a `mypy` error here. I'm looking into it, but maybe you'll know more quickly:

```
py_ecc/bls/hash_to_curve.py:115: error: Incompatible return value type (got "Tuple[optimized_bls12_381_FQ, optimized_bls12_381_FQ, optimized_bls12_381_FQ]", expected "Tuple[optimized_bls12_381_FQ2, optimized_bls12_381_FQ2, optimized_bls12_381_FQ2]")  [return-value]
```

pacrob: Ah, maybe both `map_to_curve_G1` and `clear_cofactor_G1` should use `G1Uncompressed` instead of `G2Uncompressed`?

pacrob:
Thanks for this, @petertdavies ! One question on the typing, and we'll also need some tests. If you have time, swell, otherwise I'll be looking into how to do so :grin:

petertdavies:
> Thanks for this, @petertdavies ! One question on the typing, and we'll also need some tests. If you have time, swell, otherwise I'll be looking into how to do so 😁

I've fixed the types.

We should be able to convert the G2 hash to curve tests in `tests/bls/test_hash_to_curve.py`. The test vectors are from the IETF draft, which also has G1 test vectors. I'll try to find time to look into it.

pacrob:
OK, got the tests in and docstrings cleaned up. Is there anything else that needs doing before this is good to go, @petertdavies ?

beage666:
Please help me is true

beage666 (approved):

py_ecc/optimized_bls12_381/optimized_swu.py
```diff
@@ -15,11 +16,51 @@
...
 )
 
 
+# Optimized SWU Map - FQ to G1'
+# Found in Section 4 of https://eprint.iacr.org/2019/403
+def optimized_swu_G1(t: FQ) -> Tuple[FQ, FQ, FQ]:
+    t2 = t**2
+    iso_11_z_t2 = ISO_11_Z * t2
+    temp = iso_11_z_t2 + iso_11_z_t2**2
+    denominator = -(ISO_11_A * temp)  # -a(Z * t^2 + Z^2 * t^4)
```
reedsa, line 36: Inline comments here and on line 38 are a little confusing, it doesn't seem to line up with the actual code, are these still relevant?

pacrob: I think it is. Lines 33 and 34 create `Z * t^2`. Line 35 squares the result (which is `Z^2*t^4`) and adds it to the original result. All that is then multiplied by negative `ISO_11_A`, which is the `-a`. Line 38 adds 1 to the original result, then multiplies by `b`.

beage666: Ok

reedsa: Ah thanks, hadn't noticed the other vars above line 36 were components of the whole equation. I think the inline comment threw me off but it's fine as is.

reedsa (approved):
lgtm

kclowes (approved):
LGTM! 🚢

PR #151, opened 2024-11-25
Put py38 support back in; reserve for major release cycle.

fselmo:
### What was wrong?

We opted for introducing breaking changes apart from the template updates and applying these changes only when we are ready to go through the major release cycle of all our maintained libraries.

### How was it fixed?

- Re-introduce py38 support before the next release

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

:0)


kclowes (approved):
🚀

reedsa (approved):

PR #152, opened 2025-01-13
Drop unneeded `cached-property` dependency

Xeonacid:
### What was wrong?

 `cached-property` dependency is unused.

The only usage of `cached_property` is imported from `functools`[1], not the third-party module.

[1] https://github.com/ethereum/py_ecc/blob/c13300b23671b9ccf0d339f298096ce59c9103e2/py_ecc/fields/optimized_field_elements.py#L4-L5

### How was it fixed?

Dropped the dependency from `setup.py`.

All tests passed without the dependency.

### Todo:

- [x] Clean up commit history
- [ ] Add or update documentation related to these changes

I'm not sure if a release note is needed for this change. If true, please tell me.

#### Cute Animal Picture

### DIFF
--- a/newsfragments/152.internal.rst
+++ b/newsfragments/152.internal.rst
@@ -1,0 +1,1 @@
+Remove unused ``cached-property`` dependency.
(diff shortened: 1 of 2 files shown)


pacrob:
@Xeonacid Thank you! I've added the newsfragment, looks good to go!

PR #154, opened 2025-01-30
Update LICENSE

maximevtush:
Updated the copyright year to 2025 in the LICENSE file


pacrob:
Thanks, @maximevtush, but in this case it's not necessary. We want to leave the starting year of copyright as it is. We have the convention of leaving it just that year for individuals, only adding the second year if it's an entity such as the EF.

PR #155, opened 2025-06-05
Cache subclasses of FQ

SamWilsn:
#### Cute Animal Picture

### DIFF
--- a/py_ecc/fields/field_elements.py
+++ b/py_ecc/fields/field_elements.py
@@ -12,6 +12,8 @@ from typing import (
     Union,
     cast,
 )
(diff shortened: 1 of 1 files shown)


SamWilsn:
This change cuts the running time of EELS' `stStaticCall - StaticcallToPrecompileFromCalledContract_d0g0v0_Berlin` test from 248.37s (it hit a timeout, don't know how long it actually runs for) to 5.72s, an improvement of at least 43x under PyPy 7.3.19.

There are likely more elegant ways to solve this&mdash;creating a subclass dynamically is always going to be slow in PyPy&mdash;but this works for now.

PR #156, opened 2025-07-14
Optimize ``multiply()`` and ``double()`` for point at infinity

fselmo:
### What was wrong?

Related to Issues 
- ethereum/execution-spec-tests#1894
- ethereum/execution-specs#1321

### How was it fixed?

- `double()`: The non-optimized implementation has a quick return for the point at infinity for `double()`, whereas the optimized does not have this. This was causing a slowdown for optimized vs non-optimized implementation in execution-specs when making a `subgroup_check()`.
- `multiply()`: Multiplication by infinity is infinity, quick return to skip unnecessary calculation when this is the case.

Additional changes:

- Add clarifying comments for point-at-infinity checks by checking `None` as this is required by mypy but not immediately obvious to a reader. Remove redundant `is_inf()` checks as that just checks for `None` again.
- Claude-assisted optimization for double() when y=0 was added here to be considered. I don't believe a point can be on the curve and have y=0 but maybe it's worth the quick check and return. Looking for a sanity check here.

### Todo:

- [x] Clean up commit history
- [x] ~~Add or update documentation related to these changes~~

#### Cute Animal Picture

Screenshot 2025-07-14 at 17 59 37

### DIFF
--- a/py_ecc/optimized_bls12_381/optimized_curve.py
+++ b/py_ecc/optimized_bls12_381/optimized_curve.py
@@ -94,6 +94,9 @@ if not is_on_curve(G2, b2):
 def double(
     pt: Optimized_Point3D[Optimized_Field],
 ) -> Optimized_Point3D[Optimized_Field]:
+    if is_inf(pt):
+        return pt
+
     x, y, z = pt
     W = 3 * x * x
     S = y * z
--- a/py_ecc/optimized_bls12_381/optimized_curve.py
+++ b/py_ecc/optimized_bls12_381/optimized_curve.py
@@ -18,6 +18,12 @@ curve_order = (
     52435875175126190479447740508185965837690552500527637822603658699938581184513
 )
 
+# Cache zero and one values for faster comparison and operations
+_FQ_ZERO = FQ.zero()
+_FQ_ONE = FQ.one()
+_FQ2_ZERO = FQ2.zero()
+_FQ2_ONE = FQ2.one()
+
 # Curve order should be prime
 if not pow(2, curve_order, curve_order) == 2:
     raise ValueError("Curve order is not prime")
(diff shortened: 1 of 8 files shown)


py_ecc/optimized_bls12_381/optimized_curve.py
```diff
@@ -18,6 +18,12 @@
     52435875175126190479447740508185965837690552500527637822603658699938581184513
 )
 
+# Cache zero and one values for faster comparison and operations
+_FQ_ZERO = FQ.zero()
+_FQ_ONE = FQ.one()
+_FQ2_ZERO = FQ2.zero()
+_FQ2_ONE = FQ2.one()
```
fselmo, line 25: This ends up being quite minimal so I wouldn't be opposed to leaving it out for easier readability.

py_ecc/optimized_bn128/optimized_curve.py
```diff
@@ -79,6 +79,9 @@ def is_on_curve(pt: Optimized_Point3D[Optimized_Field], b: Optimized_Field) -> b
 def double(
     pt: Optimized_Point3D[Optimized_Field],
 ) -> Optimized_Point3D[Optimized_Field]:
+    if is_inf(pt):
+        return pt
+
     x, y, z = pt
```
fselmo, line 85: I think we can also return early when `y == y.__class__.zero()` but I would need someone to verify that as no existing test hits this line when I made that change.

source: https://www.nayuki.io/page/elliptic-curve-point-addition-in-projective-coordinates

> Point doubling
...
If y=0, then 2P=O.

py_ecc/optimized_bls12_381/optimized_curve.py
kclowes: I think it's about the same readability-wise, so if it gives us a little bump I'd vote to keep it

py_ecc/optimized_bn128/optimized_curve.py
```diff
@@ -79,7 +79,15 @@ def is_on_curve(pt: Optimized_Point3D[Optimized_Field], b: Optimized_Field) -> b
...
+
+    if y == y.zero():
```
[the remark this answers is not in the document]
kclowes: I guess I'd lean toward not breaking here, but also don't think that anyone would be checking the x and y values if z=0, so I'm okay with either way you want to go.

kclowes (approved):
🚀

py_ecc/optimized_bn128/optimized_curve.py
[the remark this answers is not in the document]
fselmo: Another thing to note is I don't believe a point that is on the curve can have `y = 0`. That said, this can still serve as a safeguard and quick return.

